PR #28, opened 2016-09-16
Feature qwg

wvlothuizen:


AdriaanRol:
I'll try to take a look at it later today and leave comments where required.

AdriaanRol:
And some more things 
- [ ] basic automated tests for communications (get and set functionality) 
- [ ] better documentation of QWG and functions/parameters

wvlothuizen:
- as a first fix, I removed all code that is not relevant for the QWG (i.e. generic AWG stuff) and won't be so in the future (some functions are currently unsupported but will be in the future)
- the Transport class was a partly finished effort to split the Transport (i.e. the actual mechanism to get data to the instrument) from the object hierarchy, to allow the user to select between (say) IPInstrument and VisaInstrument, and to also formalise the interface of drivers to the Transport. I removed it, as it would require changes to QCoDeS to actually do this

AdriaanRol:
@wvlothuizen 
Another point I just ran into. 
The SCPI base class adds all the base SCPI commands but it does not do so according to the qcodes

AdriaanRol:
@wvlothuizen , Setting the sideband frequency does not work! 

`QWG.ch_pair1_sideband_frequency.set(100e6)` does not work 
`QWG.ch_pair1_sideband_frequency.get()` returns a timeout 

Running the test script initially gave no signal except for an offset (contained high and low with modulation). 
After setting the waveform to gauss I could clearly see the gaussians. 

Both these observations could be explained by a missing modulation. 
We were able to isolate the issue to the ch_pair1_sideband_frequency command not working. 

@wvlothuizen can you help with this on a short term?

AdriaanRol:
@wvlothuizen Two more commands that don't work 
`qwg1.run_mode.get()`
qwg1.trigger_level.get()` 

I know that there still are some things that are based on the 5014 but it's hard testing like this if I don't know what is and is not supposed to work

AdriaanRol:
@wvlothuizen I'm also running into communication errors. Do you think we can discuss the communication protocol in detail tomorrow? 

Best, 

Adriaan

AdriaanRol:
Hi wouter, 

Just testing the driver but there are still some bugs with the wlist. I'll try to fix them now.

pycqed/instrument_drivers/physical_instruments/QuTech_AWG_Module.py
```diff
@@ -0,0 +1,371 @@
...
+        # parameter handling
+        if len(marker1) == 0 and len(marker2) == 0:  # no marker data
+            m = np.zeros(len(waveform))
+        else:
+            if (not((len(waveform) == len(marker1))
+                    and ((len(marker1) == len(marker2))))):
+                raise UserWarning('length mismatch between markers/waveform')
+            # prepare markers
+            m = marker1 + np.multiply(marker2, 2)
+            m = int(np.round(m[i], 0))
```
AdriaanRol: @wvlothuizen 
Can you explain what this line is supposed to do? 
There are some strange things, the variable `i` in  `m[i]` is not defined anywhere, 
It is unclear to me why m = marker1 + 2*marker2 (I assume that is what that line is supposed to do). 
Converting the array to an int will also yield weird behaviour. 

Stuff I'm doing now uses no markers but only waveforms so we're still fine but I think we need to discuss this hard linking of marker and waveform channels but that is a separate discussion.

AdriaanRol:
Driver is now close to merging, points that need to be addressed: 
- [ ] marker functionality in driver. remove this from the send waveform and related commands as this is not implemented in the hardware
- [ ] test reading back of waveforms (this is not working correctly yet) 

Otherwise all points raised above (excluding docs) are addressed

Issue #38, opened 2016-09-30
QWG driver improvements


AdriaanRol:
Driver for QWG can be improved on the following points: 
- [x] marker functionality in driver. remove this from the send waveform and related commands as this is not implemented in the hardware
- [x] test reading back of waveforms (this is not working correctly yet) 

Docs also need to be written but is not a pycqed issue.

PR #47, opened 2016-10-21
Feature qwg 2

wvlothuizen:
The most notable line of code (in SCPI.py) is:

```
        self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
```

This speeds up the tests in QWG_tests from 5.350 s to 0.465 s. It might be a great addition to IPInstrument for other drivers suffering slow response times


AdriaanRol:
Closes #38 . Merging as the issues are addressed.

PR #50, opened 2016-11-04
Added QX Client Interface

Nader-Khammassi:
Added QX Sweep Function
Added QX Detector Function


AdriaanRol:
@Nader-Khammassi Looks pretty awesome. I will merge as soon as the tests pass. 

We also talked about your example ipython notebook. Do you think it is possible to add (possibly in a new PR) an self-contained example notebook that also explains the requirements and how to set up the QX simulator server.

PR #53, opened 2016-11-04
Fixpoint problem

AdriaanRol:
This should fix the fixpoint problems !


AdriaanRol:
Pull request also contains the recent history of the 2Q branch (some untested stuff as well but hey).

PR #59, opened 2016-11-17
Soft averaging

AdriaanRol:
This solves #58 . 

Soft averaging has been added for 
- 1D hard sweeps 
- nD hard sweeps

The number of soft averages is a parameter of the MC and remains set for all hard sweeps once it is set. 
It can be set using `MC.soft_avg(N)` 

Additionally it 
- repairs the missing progress print statement that got broken at some point 
- converts  the MC to an instrument (this does not change anything for the user) 

- Adds more tests for the measurement control 
- Added hard sweeps and soft averages to the example notebook 
- general cleanup of MC 

@elrama- @NielsBultink , Please take a look and merge if you agree with these changes.


elrama-:
Tested in my computer and works

PR #63, opened 2016-11-19
Fridge monitor 

### LINKED ISSUE #60: Fridge temperature monitor instrument
I just ported over the python 2 version of the fridge monitor in c4b594409998 . 

I want to improve this driver a bit to be up to PycQED_py3/QCoDeS standards and then merge it into the master.

AdriaanRol:
Closes #60 
@elrama- could you take a look at this PR and merge it if OK or suggest changes if it is not. 

This is a port of the fridge monitor to pycqed_py3/qcodes. 

I cleaned up the code a bit and made it more robust against failing. 
This fridge monitor is based on the version Serwan made for the old pycqed. 

Some things to note: 
- Every temperature being monitored is it's own parameter. 
- When a value is not a double on the website but e.g., 0 (and int) or inf (no reading) reading that specific parameter fails and a warning is displayed and the parameter is set to 0. 
- sometimes reading values from the website fails, this will trigger a warning and will not update the internal dictionary or try again. 
- I overwrote snapshot so a normal measurement will correctly update the values once every update interval. 

Areas where this driver could be improved (that I will not do as part of this PR):
- Make it more robust against failures in the website reading (by e.g., retrying) 
- Make it correctly read non double values (integers) 
- Make it determine the parameters and units to read based on what it reads on the website. 
- Include pressures in the fridge mon instrument (should not be too hard). 

There is an example notebook you can run. Maybe you need to run it a few times as the website comms fail sometimes.


AdriaanRol:
This branch is currently in a working state. This means that the fridge monitor correctly logs the fridge data and does not give hard crashes. This means that the current version closes #60 

However it does still fail to read out sometimes (roughly once every few times) raising a warning. It is not currently my top priority to make this more robust. As such I propose we merge this ( @elrama- , @NielsBultink, @quantumCryoCoder?) and open a new issue for improving this.

Issue #61, opened 2016-11-18
Soft Avg not working in spectroscopy


elrama-:
I pulled the changes from PR #59 into 2Q_gates_branch (e953d78). Tried to use it for continuous resonator measurements, but stops after the first iteration. However, MC is prediting the time correctly, it is just exiting the loop prematurely.

AdriaanRol:
@elrama- I have implemented soft averaging for hard-sweeps. I did not think of also implementing it for soft sweeps but I can see the use (fast spec scans and then repeating till the signal is clear sounds like a very pleasant way of working).

PR #64, opened 2016-11-20
Soft averaging for soft sweeps 

AdriaanRol:
@elrama- As requested in #61 I have now implemented soft averaging for soft sweeps. This works both for 1D, 2D and N-D. 

This pull request contains 
* Updated example notebook, including some options for the MC and adaptive sweeps 
* new tests for adaptive measurements as well as for soft averaging for soft sweeps. 

Check it out and let's merge it!

### DIFF
--- a/examples/Setting up a basic measurement using the MC.ipynb
+++ b/examples/Setting up a basic measurement using the MC.ipynb
@@ -429,131 +429,10 @@
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


AdriaanRol:
Using it to measure Rabi's on the setup now. Works so I'm merging this into the master. 
Closes #61

PR #70, opened 2016-11-28
Prettier plotting

### LINKED ISSUE #69: Prettify MC live plot
The QCoDeS live plot has prettier lines than the default version I use. For the demo I could try to update our plotting. 

This is a low priority fix.

AdriaanRol:
Closes #69 
Changes in this pull request 
* replaces my custom build plot-monitor with the QCoDeS plot monitor (this was previously used only for the 2D plotting). 
* Adds a "plotting_interval" parameter to the MC that determines how often the live plot updates

Below is a screenshot of what the updated plotting looks like. 
image

### DIFF
--- a/examples/Setting up a basic measurement using the MC.ipynb
+++ b/examples/Setting up a basic measurement using the MC.ipynb
@@ -389,8 +389,6 @@
     }
    ],
    "source": [
-    "\n",
-    "\n",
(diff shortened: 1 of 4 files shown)


AdriaanRol:
Note: not ready for merging as the later added persist mode still has some minor problems, which I'm working on now.

AdriaanRol:
Added tests and fixed bugs. We are testing it now in the demo. 
Would like someone else to take a look at it before merging.

PR #78, opened 2016-12-14
Added the git flowchart as per Flo's request

AdriaanRol:


AdriaanRol:
@fluthi , Can you check out if this solves your problem? (if OK, hit the review button and click OK).

PR #84, opened 2016-12-22
Added n qubit general linear and MLE  tomography toolbox

GitJaap:
Qutip based code for state tomography based on an arbitrary order of applied clifford gates and arbitrary number of qubits.

### DIFF
--- a/pycqed/analysis/tomography.py
+++ b/pycqed/analysis/tomography.py
@@ -1,0 +1,248 @@
+import qutip as qtp
+import numpy as np
+import time
+import scipy
+
+
+class TomoAnalysis():
+
+    """Performs state tomography based on an overcomplete set of measurements
+     and calibration measurements. Uses qutip to calculate resulting basis states
+     from applied rotations
+
+    Uses binary counting as general guideline in ordering states. Calculates
+     rotations by using the qutip library
+    BEFORE YOU USE THIS SET THE CORRECT ORDER BY CHANGING
+        'rotation_matrixes'
+        'measurement_basis' + 'measurement_basis_labels'
+        to values corresponding to your experiment
+        and maybe 'readout_basis'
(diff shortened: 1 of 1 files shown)


AdriaanRol:
Cool, I'll probably take a close look at this during the christmas holidays. Good work :+1:

elrama-:
Code works, Chris' project already shows that. I just changed the naming to denote that this code is implemented for a joint RO (only one set of betas). After checks are passed I will merge

AdriaanRol:
@elrama- , did you test this? I'm mostly talking about potential python 3 vs python2 conversion bugs. I trust the logic is fine (as you helped and @cdickel  is using it).

AdriaanRol:
Also @elrama- , what do you mean with a joint RO? I'm assuming it is a simultaneous readout (similar to a multiplexed RO?). If it is just that I suggest we drop the rename. If it is something else, we may want to clarify that in the docstrings.

elrama-:
@AdriaanRol 
> Did I test it? I am on that right now.
> A Joint RO is a special kind of RO, that requires a special kind of RO operator reconstruction, since it naturally contains information about correlations. For a multiplexed RO, we have to produce this information by multiplying shot by shot.

These two cases: 1. Multiplexed RO(Ideally 2 uncorrelated channels) and 2. Joint RO (One correlated channel), have to be treated differently as the second involves matrices of bigger dimensions.

> About docstrings, that is preciselly what I added in my commit, line 13 in the docstring that reads "Works for Joint RO (like REM experiment)."

> About naming, the used names in literature[1] are those: Joint RO and Multiplexed RO.

[1] Detecting highly entangled states with a joint qubit readout. Chow, J. M. et al. http://journals.aps.org/pra/abstract/10.1103/PhysRevA.81.062325

PR #90, opened 2017-01-16
Updates on Tomo

elrama-:
There were minor updates on the Tomo code that require this pull request. Main ones being:

> Update on MLE speed (NICE FEATURE)
> dimension-matching of qutip Qobj. (REQUIRED)


AdriaanRol:
@elrama- , As I have to do tomo analysis today, I will convert your notebook analysis to a measurement analysis function. I'll add this to this branch so that we can include it in a PR. I'll also add a sample dataset as a test. 
- [ ] Combine tomo in MA script 
- [ ] sample dataset with test for analysis

AdriaanRol:
More enhancements for tomo (probably not in this PR but rather when I'm needing them). 

calculate interesting metrics 
- CHSH 
- concurrence

PR #98, opened 2017-01-22
Enabled accessing instruction execution state

gtaifu:
- Enabled reading back the max instruction address that has been
executed by the QuMA processor. This is done by adding the parameter
*max_instruction_address*.
- Enabled accessing the QuMA state indicating if the pulse queues has
gone empty (i.e. the instruction issue rate is slower than the
consumption rate).  This is done by adding the parameter
*pulse_queue_state*.

CBox_v3 starts to support the feature since [version 3.3.3](https://github.com/DiCarloLab-Delft/250MSps_ControlBox/releases/tag/code_v3.3.3).

@AdriaanRol Please check if the parameter names are good.


pycqed/instrument_drivers/physical_instruments/_controlbox/defHeaders_CBox_v3.py
```diff
@@ -27,3 +27,6 @@
 trigger_sources = ['0: internal',
                    '1: external',
                    '2: mixed']
+
+pulse_queue_states = ['0: normal',
+                      '1: error']
```
AdriaanRol: Is there any way this error can be more explicit? (I guess not).

pycqed/instrument_drivers/physical_instruments/QuTech_ControlBox_v3.py
```diff
@@ -38,6 +38,15 @@ def add_params(self):
...
                            vals=vals.Enum('internal', 'external',
                                           'mixed'))
+        self._pulse_queue_state = defHeaders.pulse_queue_states[0]
+        self.add_parameter('pulse_queue_state',
+                           get_cmd=self._do_get_pulse_queue_state)
+        self._max_instruction_address = 0
+        self.add_parameter('max_instruction_address',
+                           get_cmd=self._do_get_max_instruction_address)
+        self._uploaded_program_length = 0
+        self.add_parameter('uploaded_program_length',
```
AdriaanRol: Good parameter name :+1:

pycqed/instrument_drivers/physical_instruments/QuTech_ControlBox_v3.py
```diff
@@ -38,6 +38,15 @@ def add_params(self):
                            get_cmd=self._do_get_trigger_source,
                            vals=vals.Enum('internal', 'external',
                                           'mixed'))
+        self._pulse_queue_state = defHeaders.pulse_queue_states[0]
+        self.add_parameter('pulse_queue_state',
```
AdriaanRol: This parameter may benefit from a docstring, explaining what this paramter means. I have some sort of idea but it is rather vague. I think that the name (depending on what it does) is still a good name. 

A docstring can be added quite simply by using "docstring = ". I've added a little snippet from another instrument with an example of how to do that. 

```python 
        self.add_parameter('kernel_list',
                           initial_value=[],
                           vals=vals.Anything(),
                           parameter_class=ConfigParameter,
                           docstring='List of external kernels to be loaded')
```

pycqed/instrument_drivers/physical_instruments/QuTech_ControlBox_v3.py
```diff
@@ -38,6 +38,15 @@ def add_params(self):
...
+                           get_cmd=self._do_get_pulse_queue_state)
+        self._max_instruction_address = 0
+        self.add_parameter('max_instruction_address',
```
AdriaanRol: I understand what this means because you explained it to me. Again a dosctring would be good.

AdriaanRol:
@gtaifu , I want to start using the content of this branch after Maserati cools down but I left quite a few comments on this PR. Do you have any plan to address these issues?

pycqed/instrument_drivers/physical_instruments/_controlbox/defHeaders_CBox_v3.py
gtaifu: Currently not.

pycqed/instrument_drivers/physical_instruments/QuTech_ControlBox_v3.py
gtaifu: I will add the docstring.

pycqed/instrument_drivers/physical_instruments/QuTech_ControlBoxdriver.py
```diff
@@ -94,7 +94,7 @@ def add_params(self):
                 label='Integraion weights input {}'.format(i),
                 get_cmd=self._wrap_ch_get_fun(self._get_int_weights, i),
                 set_cmd=self._wrap_ch_set_fun(self._set_int_weights, i),
-                vals=vals.Anything())
+                vals=vals.Ints(-128, 127))
```
AdriaanRol: This will not work. It will check if the weights is an Integer where it should be an Array. 

Try using the Arrays validator (this allows both specifying a min and maxvalue as well as a shape. Maybe you can even make a custom one that also checks type. 

There is also a good reason to use validators as they have auto generated docstrings so that the user can easily see what input is expected.

PR #105, opened 2017-01-27
modifications to analog output level and number of codewords to match…

wvlothuizen:
… QWG software 0.9beta4


AdriaanRol:
I like it, very minimal changes to the driver! 
Is this ready to merge?

PR #110, opened 2017-02-02
First try at improving the latency of the UHFLI-QC driver (for Niels Bultink)

nielshaandbaek:
There are a number of changes proposed here. First of all, the driver now uses only a single instance of the AWG module. This is done in order to reduce memory usage when the AWG is configured multiple times. In addition, the readout functions have been changed to avoid subscribing/unsubscribing all the time as that also wastes time.

To test the latency of the new driver use something like this:

UHFQC_1.single_acquisition_initialize(set([0]), 'rl')

delays = []
for n in range(0, 10):
    t0 = time.time()
    data = UHFQC_1.single_acquisition_poll(RESULT_LENGTH, 0.0010, 1000)
    delays.append(1e3*(time.time()-t0))

UHFQC_1.single_acquisition_finalize()

print("Total time per iteration: {:.0f} +/- {:.0f} ms".format(np.mean(delays), np.std(delays)))


AdriaanRol:
@NielsBultink do you have time to look into this before you fly to the US?

NielsBultink:
Just testing before leaving to the US and getting an error message:
image

NielsBultink:
It seems zhinst.ziPython in ziPython3.5_ucs4-16.12.41424-win64.msi"does not have the required functions:
 asyncSetInt()

nielshaandbaek:
Sorry about that. I have uploaded a new version of the Python API to the website that should include these functions. https://people.zhinst.com/~niels/

In addition, the Github messages got sorted into my Junk folder by mistake, which is why I've been a bit slow to react. That should also be fixed by now.

PR #116, opened 2017-02-20
Ddm branch

NikitaVodyagin:
First pull request for the code review of DDM SW


pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
+'''
```
AdriaanRol: What is the difference between the DDM module and the DDMq? 

This is quite unclear to me, which one do I need to review?

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+class DDMq(SCPIddm):
+
+    # def __init__(self, logging=True, simMode=False, paranoid=False):
+    def __init__(self, name, address, port, **kwargs):
+        super().__init__(name, address, port, **kwargs)
+        # AWG properties (to be overloaded by derived class)
+        self.device_descriptor = type('', (), {})()
+        self.device_descriptor.model = 'DDM'
+        self.device_descriptor.numChannels = 2
+        # self.numDacBits = 0
```
AdriaanRol: Try to delete commented out lines.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+        #######################################################################
+        # DDM specific
+        #######################################################################
+
+        for i in range(self.device_descriptor.numChannels//2):
+            ch_pair = 2*i+1
+            snavg_cmd = 'qutech:inputavg{}:naverages'.format(ch_pair)
+            self.add_parameter('ch_pair{}_inavg_Navg'.format(ch_pair),
+                               units='#',
+                               label=('number of averages' +
```
AdriaanRol: This label will have a missing space between number of averages and ch_pair. 

Also, let's start Labels with capital letters.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               set_cmd=snsamp_cmd + ' {}',
+                               vals=vals.Numbers(1, 4096)
+                               )
+            senable_cmd = 'qutech:inputavg{}:enable'.format(ch_pair)
+            self.add_parameter('ch_pair{}_inavg_enable'.format(ch_pair),
+                               label=('enable input averaging' +
+                                      'chpair {} '.format(ch_pair)),
+                               get_cmd=senable_cmd + '?',
+                               set_cmd=senable_cmd
+                               # vals=vals.Numbers(1,4096)
```
AdriaanRol: I guess we do want to have some kind of validator here?

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               )
+            sdisable_cmd = 'qutech:inputavg{}:disable'.format(ch_pair)
+            self.add_parameter('ch_pair{}_inavg_disable'.format(ch_pair),
+                               label=('disable input averaging' +
+                                      'ch_pair {} '.format(ch_pair)),
+                               get_cmd=senable_cmd + '?',
+                               set_cmd=sdisable_cmd,
+                               # vals=vals.Bool()
```
AdriaanRol: again, missing validator

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+            self.add_parameter('ch_pair{}_rotmat_rotmat11'.format(ch_pair),
+                               units='#',
+                               label=('rotation matrix value 11 of' +
+                                      'ch_pair {} '.format(ch_pair)),
+                               get_cmd=srotmat11_cmd + '?',
+                               set_cmd=srotmat11_cmd + ' {}'
+                               # vals=vals.Numbers(1,2048)
+                               )
+            '''
+            TV mode parameters
```
AdriaanRol: In the existing drivers `TV mode` is called `Integration averaging` or `int_avg`. To be consistent this should be renamed to integration averaging. (also in everywhere following this. 

Of course this does not require changing the VISA commands.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+            ch = i+1
+            self.add_parameter('ch{}_inavg_data'.format(ch),
+                               label=('get data ch {} '.format(ch)),
+                               get_cmd=self._gen_ch_get_func(
+                                   self.getInputAverage, ch),
+
+                               vals=vals.Numbers(-128, 127)
+                               )
+            srotres_cmd = 'qutech:wintrot{}:result'.format(ch)
+            self.add_parameter('ch{}_wintrot_result'.format(ch),
```
AdriaanRol: the name `wintrot_result` is a bit confusing for me. Do you think we can have a better name for this? 

Also in this case a docstring would be helpful.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               units='#',
+                               label=('rotated integration result of' +
+                                      'channel {} '.format(ch)),
+                               get_cmd=srotres_cmd + '?'
+                               # vals=vals.Numbers(-134217728,134217727)
+                               )
+            sthl_cmd = 'qutech:qstate{}:threshold'.format(ch)
+            self.add_parameter('ch{}_qstate_threshold'.format(ch),
+                               units='#',
```
AdriaanRol: add a docstring that explains what the threshold is used for/when it is used.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               label=('set threshold of ch {} '.format(ch)),
+                               get_cmd=sthl_cmd + '?',
+                               set_cmd=sthl_cmd + ' {}'
+                               # vals=vals.Numbers(-134217728,134217727)
+                               )
+            self.add_parameter('ch{}_weight_data'.format(ch),
```
AdriaanRol: I assume these are the integration weights? 

I propose shorting the parameter name to `ch{}_weights` , and clarifying what weights they are in the label and docstring.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               label=('get weigtht data ch {} '.format(ch)),
```
AdriaanRol: This label does not describe what data this is, to clarify, if we would plot this we would not put this on the axis label. 
I think a better label would be : `Integration weigths channel {}`

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               get_cmd=self._gen_ch_get_func(
+                                   self.getWeightData, ch),
+                               set_cmd=self._gen_ch_set_func(
+                                   self.sendWeightData, ch),
+                               vals=vals.Arrays(-128, 127)
```
AdriaanRol: The validator contains constraints on the values but not on the shape. Maybe you want to specify a max shape? 

I'm not sure if the array validator supports this by default (I made it so we can change that) but I think it should give an error if we put in an array that is too large.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               )
+
+    def getInputAverage(self, ch):
```
AdriaanRol: This should be a private method right? (starting with `_`).

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+
+        self.write('qutech:inputavg{:d}:data? '.format(ch))
+        binBlock = self.binBlockRead()
+        print('inputavgLen in bytes {:d}'.format(len(binBlock)))
+        inputavg = np.frombuffer(binBlock, dtype=np.float32)
+        return inputavg
+
+    def getTVdata(self, ch_pair):
```
AdriaanRol: This should also be private right ? 
Also I would replace this with get_int_avg_data

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               )
+            self.add_parameter('ch_pair{}_tvmode_data'.format(ch_pair),
+                               label=('get TV data ch {} '.format(ch_pair)),
+                               get_cmd=self._gen_ch_get_func(
+                                   self.getTVdata, ch_pair)
+
+                               # vals=vals.Numbers(-128,127)
+                               )
+            scaladc_cmd = 'qutech:caladc{}'.format(ch_pair)
+            self.add_parameter('ch_pair{}_cal_adc'.format(ch_pair),
```
AdriaanRol: From what I can see, this is not a parameter (it's not getable) but rather a function. Calibrate ADC. 

I don't know what it does so maybe it could be a parameter but it doesn't look like it.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               label=('Calibrate ADC{}'.format(ch_pair)),
```
AdriaanRol: Similarly this would better fit docstring than label.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+            stven_cmd = 'qutech:tvmode{}:enable'.format(ch_pair)
+            self.add_parameter('ch_pair{}_tvmode_enable'.format(ch_pair),
+                               label=('enable tv mode' +
+                                      'ch_pair {} '.format(ch_pair)),
+                               get_cmd=stven_cmd + '?',
+                               set_cmd=stven_cmd
+                               # vals=vals.Numbers(1,4096)
+                               )
+            stvdis_cmd = 'qutech:tvmode{}:disable'.format(ch_pair)
+            self.add_parameter('ch_pair{}_tvmode_disable'.format(ch_pair),
```
AdriaanRol: The way this parameter is named, it is a function. 

I would say that the corresponding parameter would be 
`ch_pair{}_tvmode_state` (or even better `ch_pair{}_int_avg_enabled`). 

This would then be getable (is the state enabled or disabled?) and setable, you can turn it on and off. 

Right now this is a function instead of a parameter. It should be a parameter.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               label=('the number of integration averages' +
+                                      'of ch_pair {} '.format(ch_pair)),
+                               get_cmd=sintavg_cmd + '?',
+                               set_cmd=sintavg_cmd + ' {}',
+                               vals=vals.Numbers(1, 1024)
+                               )
+            stvseg_cmd = 'qutech:tvmode{}:nsegments'.format(ch_pair)
+            self.add_parameter('ch_pair{}_tvmode_nsegments'.format(ch_pair),
+                               units='#',
+                               label=('the number of TV segments of' +
```
AdriaanRol: I propose `Number of segments ch_pair{}`. 

Also, does this mean I can have a per channel setable number of segments? We currently only do this globally (for all channels at once) but I kind of like this extra freedom.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+                               set_cmd=srotmat11_cmd + ' {}'
+                               # vals=vals.Numbers(1,2048)
+                               )
+            '''
+            TV mode parameters
+            '''
+            sintavg_cmd = 'qutech:tvmode{}:naverages'.format(ch_pair)
+            self.add_parameter('ch_pair{}_tvmode_naverages'.format(ch_pair),
+                               units='#',
+                               label=('the number of integration averages' +
```
AdriaanRol: `Number of integration averages ch_pair{}`

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+            swintstat_cmd = 'qutech:wint{}:status'.format(ch_pair)
+            self.add_parameter('ch_pair{}_wint_status'.format(ch_pair),
+                               units='#',
+                               label=('weighted integral status of' +
+                                      'ch_pair {} '.format(ch_pair)),
+                               get_cmd=swintstat_cmd + '?'
+                               # vals=vals.Numbers(1,2048)
+                               )
+            srotmat00_cmd = 'qutech:rotmat{}:rotmat00'.format(ch_pair)
+            self.add_parameter('ch_pair{}_rotmat_rotmat00'.format(ch_pair),
```
AdriaanRol: I would propose 
`ch_pair{}_rotmat_00` to make the parameter name a bit shorter

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+from qcodes import validators as vals
+
+
+class DDMq(SCPIddm):
+
+    # def __init__(self, logging=True, simMode=False, paranoid=False):
+    def __init__(self, name, address, port, **kwargs):
+        super().__init__(name, address, port, **kwargs)
+        # AWG properties (to be overloaded by derived class)
+        self.device_descriptor = type('', (), {})()
```
AdriaanRol: AFAIK these device descriptors and such are note used. 
Try commenting it out and testing if it still works. No sense in keeping artefacts of Wouters QWG driver.

pycqed/instrument_drivers/physical_instruments/DDM/DDMq.py
```diff
@@ -0,0 +1,318 @@
...
+            snsamp_cmd = 'qutech:inputavg{}:scansize'.format(ch_pair)
+            self.add_parameter('ch_pair{}_inavg_scansize'.format(ch_pair),
+                               units='#',
+                               label=('number of pair samples' +
+                                      'ch_pair {} '.format(ch_pair)),
+                               get_cmd=snsamp_cmd + '?',
+                               set_cmd=snsamp_cmd + ' {}',
+                               vals=vals.Numbers(1, 4096)
+                               )
+            senable_cmd = 'qutech:inputavg{}:enable'.format(ch_pair)
```
AdriaanRol: I'm not sure what senable means but the variable name could be a bit clearer (but this is minor).

AdriaanRol:
And another small request, we may want to come up with a shorthand for `ch_pair` as we'll be typing that a lot. What do you think about `chp`? The label and docstring should still clarify that this is about channel pair. 

Also @NielsBultink what do you think about this?

PR #117, opened 2017-02-21
Updates to support version 5 of the UHFLI-QC firmware.

nielshaandbaek:
This pull request needs the features associated with version 5 of the UHFLI-QC firmware and the corresponding LabOne software (see https://people.zhinst.com/~niels/).

The pull request adds some new nodes to the s_node_pars.txt file. The file could be regenerated but I found it easier to just modify the file.

In case of the driver itself, I modified it to _not_ use the autoconnect functionality. The reason is that this can really screw with the configuration in case more than one instrument is connected to the server. Now, it will only connect to the instrument that is actually selected.

In addition, the function for programming the AWG was updated to better detect when the AWG is actually ready for use. This is done using the new 'awgs/0/ready' node. It goes to '1' when the AWG is ready to execute a configured program.


AdriaanRol:
@NielsBultink I understand you have merged this into some other branch because of conflicts. 

If this is true we should be able to safely delete this branch and close the PR, is this correct? 

If so, can you close the PR (and preferably open a new one containing the changes).

AdriaanRol:
@NielsBultink . I understood this can be deleted as you have merged it into another branch. Can you confirm this?

NielsBultink:
yes, correct. Because of the merge confluct I manually incorporated the
files in thge 2Q_gates_branch.
Best, Niels

2017-03-06 11:10 GMT+01:00 Adriaan <notifications@github.com>:

PR #134, opened 2017-03-03
Psd analysis methods

elrama-:
This function implements a standardized analysis of noise properties from TimeDomain Arches scan of the qubits.

The plots generated are:

coherence_times
gamma_ratios
gamma_fit


pycqed/analysis/PSD/standard_arches_psd.py
```diff
@@ -0,0 +1,333 @@
...
+    table[3, :len(T2_star)] = T2_star
+    table[4, :len(T2_echo)] = T2_echo
+    table[5, :len(T1_mask)] = np.logical_or(np.logical_or(T1_mask,
+                                                          T2_star_mask),
+                                            T2_echo_mask)
+
+    return table
+
+
+def arch(dac, Ec, Ej, offset, dac0):
```
AdriaanRol: Should use the arch as defined in the fit_mods, that should be updated to be consistent with this as per #143 (N.B. this version is correct).

PR #147, opened 2017-03-20
Live monitor

### LINKED ISSUE #141: Safe interrupt functionality for MC
In PycQED py2 we had a safe interupt for the MC. When pressing `q` during a measurement it would stop the experiment on the next iteration. The advantage is that this can not accidentally crash the kernel, as pressing the stop button in the notebook or pressing ctrl+c on the command line can.

The reason that this was removed was that the notebook does not support detecting keystrokes. However, since #126 this is possible again.

AdriaanRol:
A unit aware live instrument monitor. 

To use it add the following to your init: 

```python 
from pycqed.instrument_drivers.virtual_instruments import instrument_monitor as im
IM = im.InstrumentMonitor('IM', station)
station.add_component(IM)
MC.instrument_monitor(IM.name)
```

The monitor has the following features 
- updates every few (default 5) seconds triggered from the MC
- unit awareness. If the unit is an SI unit it will apply SI prefixes and rescale the value (e.g., Hz -> GHz or MHz) 
- instruments are collapsable so you don't have to see all the clutter all the time. 

It still has some issues that can be improved. I think these are low priority and will not do them for this PR. These are
- Everytime you update the instrument monitor it folds the tree structure again 
- Arrays and dicts can be displayed in a nicer way 
- It could be a True monitor that does not need to be triggered. (currently not allowed by qcodes architecture). 

## Example of what the monitor looks like. 
image

Let me know what you guys think. 
Also @NielsBultink can you test this out and report anything that's wrong to me so we can fix before merging. 

This PR also contains a reintroduction of q as the interrupt button. Be aware that it requires a terminal to function. This closes #141 ( @elrama- I think you'll like this).


AdriaanRol:
Just thought of a nice enhancement for this monitor. 
- add the type of the instrument besides the name. Particularly useful for instruments of which we have multiple variants such as qubits and MW sources.

PR #157, opened 2017-03-28
Update the CBox assembler to hack the hardware bug

gtaifu:
This patch is made for this bug [CBox issue #75](https://github.com/DiCarloLab-Delft/250MSps_ControlBox/issues/75)

- Added one nop instruction following each label.
- Append 4 nop instructions following beq/ben instructions.
- Add getTextInstructions() function to show the final instructions after expansion.
- Make test suite for the assembler.

Note: this is not the solution to the bug, but a hack to enable Adriaan to work without being disturbed so much.


pycqed/instrument_drivers/physical_instruments/_controlbox/Assembler.py
```diff
@@ -567,8 +570,65 @@ def convert_to_instructions(self):
...
-        instructions.append(int('a08203e8', 16))
-        instructions.append(int('12007ffd', 16))
+        self.instructions.append(int('c08003e8', 16))
+        self.instructions.append(int('a08203e8', 16))
+        self.instructions.append(int('12007ffd', 16))
+
+        return self.instructions
+
+    # used to print the expanded version of instructions uploaded
+    def getTextInstructions(self):
```
AdriaanRol: this need a docstring

AdriaanRol:
Will test today and merge if OK

AdriaanRol:
Tested on setup and working as intended. Will merge after test suites pass.

PR #164, opened 2017-04-04
Improved wait instruction assembler

### LINKED ISSUE #163: CBox assembler request
If a `wait 0` instruction is part of a qumis file. Strip the instruction, if `wait ` is called with a negative number raise an error. 

This is relevant for doing e.g. Ramsey experiments.

lciorciaro:
Assembler now ignores 'wait 0' instructions and raises a ValueError for wait instructions with negative parameter.
Added test cases with 'wait 0' and negative wait instructions.

This closes #163 . Still needs to be tested on hardware

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/_controlbox/Assembler.py
+++ b/pycqed/instrument_drivers/physical_instruments/_controlbox/Assembler.py
@@ -514,7 +514,15 @@ class Assembler():
                 self.instructions.append(int(self.MeasureFormat(), 2))
 
             elif (elements[0].lower() == 'wait'):   # Wait imm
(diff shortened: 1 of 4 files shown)


AdriaanRol:
@gtaifu , You said that you preferred us not to mess with the Assembler. However, I would really like to resolve this issue (#163) as not having this breaks some of our automated calibration routines. 

As this should be a simple python only fix I would really like it if you could prioritize this.

gtaifu:
The change in the assembler is buggy. For this part, I will rewrite it.
The test_wait_zero and test_wait_negative will be merged into the branch I am working on.

PR #165, opened 2017-04-05
Spsa optimization

elrama-:
Implemented the SPSA algorithm for the adaptive control of MC

Implemented the tests for this algorithm

For future: After acquiring more experience, decide on optimal values for the gain parameters

### DIFF
--- a/pycqed/measurement/optimization.py
+++ b/pycqed/measurement/optimization.py
@@ -156,14 +156,14 @@ def SPSA(fun, x0,
             no_improv_thr
         maxiter (int): always break after this number of iterations.
             Set it to 0 to loop indefinitely.
-        alpha, gamma, a, c, A: parameters for the SPSA gains 
+        alpha, gamma, a, c, A: parameters for the SPSA gains
             (see refs for definitions)
         p: probability to get 1 in Bernoulli +/- 1 distribution
             (see refs for context)
     return: tuple (best parameter array, best score)
 
     alpha, gamma, a, c, A and p, are parameters for the algorithm.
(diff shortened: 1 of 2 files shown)


AdriaanRol:
@elrama- Picture? (e.g., the 2 parameter example)

elrama-:
image

both colors and gray (previous run) are working examples, the difference is in the gain parameters of the algorithm (the ones that would be nice to know more about to do nice initial guesses).

The test implemented in tests_MC is the color line.

pycqed/measurement/optimization.py
```diff
@@ -135,3 +135,99 @@ def nelder_mead(fun, x0,
...
+            and operate over a numpy array of the same dimensions as x0
+        x0 (numpy array): initial position
+
+
+        no_improv_thr,  no_improv_break (float, int): break after
+            no_improv_break iterations with an improvement lower than
+            no_improv_thr
+        maxiter (int): always break after this number of iterations.
+            Set it to 0 to loop indefinitely.
+        alpha, gamma, a, c, A: parameters for the SPSA gains
```
AdriaanRol: @elrama- could you add the expected type ('m assuming float) to the docstring of these args? (better to be consistent with the rest.

elrama-: done

AdriaanRol:
@elrama- I was about to merge this but noticed you added some extra parameters that I don't know what they do and you commented out some piece of code. Could you add what they do to the docstring (or refer the source of the docs as in the rest?) then it would be absolutely perfect!

PR #169, opened 2017-04-06
Changed the assembler to a layered structure

gtaifu:
Since the instructions are read from the file, every step can perform only one functionality. An intermediate representation of the QuMIS is used. 

Added more test for the assembler.
The new assembler has the same result to the old assembler for all .qumis  files in the directory: pycqed\tests\test_data\20170328\.


AdriaanRol:
@gtaifu , Both tests are failing (you can click on it to get a detailed look). The travis build seems to fail on some instr_label parameter in the assembler (looks like a very real thing to me). 

Codacy is complaining about some more stylistic and error prone things. I've gone over some of them and there are a few that seems to indacate a real error. Can you fix these things and let me know when it is ready to review/test?

gtaifu:
I have resolved the conflict with the master branch by using my related code for QuMIS assembler.

AdriaanRol:
@gtaifu Tested and working. Will merge as soon as checks pass from merging master in here.

Issue #139, opened 2017-03-08
Include "robust" ramsey find frequency as an option in find_frequency


AdriaanRol:
We ( @elrama-  and me)  have found the following snippet very useful. It should be integrated in the qubit object. 
```python 
    times = np.arange(0, .1e-6, 1e-9)
    artificial_detuning = 2/times[-1]

    for i in range(2):
        QR.measure_ramsey(times, artificial_detuning=artificial_detuning)
        a = ma.Ramsey_Analysis(auto=True, close_fig=True)
        fitted_freq = a.fit_res.params['frequency'].value
        detuning = fitted_freq-artificial_detuning
        QR.f_qubit(QR.f_qubit()-detuning)
```

fluthi:
Quick fix: as strongly detuned pulses might yield very large amplitudes (calibration points are far off), it is possible to relax the fit constraints to ensure a convergence of the fitting procedure.

MiguelSMoreira:
Closed due to inactivity

PR #182, opened 2017-04-11
Bugfix to parabola analysis and enhancement of ramsey

fluthi:
This relates to #139 but does not close the issue.


AdriaanRol:
I think this fix looks good as is but would be better if we add a simple test using one of these "outlier" datasets for the Ramsey. 

I've asked @fluthi for some datasets we can test against. I'll add these to this PR before merging.

Issue #185, opened 2017-04-11
find_amplitude does not necessarily take best quadrature for fitting


fluthi:
In case the information is mostly in one quadrature, and the amplitudes for the first iteration are not covering a full period (although in the present case both minima are visible) the procedure does not necessarily use the quadrature with the most information.
The way it is implemented at the moment is that only the standard deviation of the fitted periods of the quadratures are taken into account to decide what amplitude to use.
Fix: take quadrature where the residuals of the fit are smallest, and use the amplitude extracted from there.

PR #190, opened 2017-04-13
Bugfix/#185/rabi analysis impovements

fluthi:
Manly addressing Rabi analysis and tests for the Rabi analysis, adresses issue #185


AdriaanRol:
@fluthi Hi Flo, it looks like some syntax error snuck in (see the codacy review details). Looks like a simple typo. 
It also looks like the entire find pulse amplitude is still commented out in this version. Could you fix that?

pycqed/analysis/measurement_analysis.py
```diff
@@ -1190,9 +1190,19 @@ def fit_data(self, print_fit_results=False, **kw):
         self.fit_res = ['', '']
         # It would be best to do 1 fit to both datasets but since it is
         # easier to do just one fit we stick to that.
+        # We make an initial guess of the Rabi period using both quadratures
+        data = np.sqrt(self.measured_values[0]**2+self.measured_values[1]**2)
```
AdriaanRol: I guess the underlying assumption here is that standard IQ demodulation is used such that this will not amplify noise in one channel more than in the other. I think that assumption is fine at this time.

pycqed/instrument_drivers/meta_instrument/qubit_objects/qubit_object.py
```diff
@@ -386,32 +386,37 @@ def find_frequency(self, method='spectroscopy', pulsed=False,
...
             MC = self.MC
         if np.size(amps) != 1:
             self.measure_rabi(amps, n=1, MC=MC, analyze=False)
             a = ma.Rabi_Analysis(close_fig=close_fig)
+            # Decide which quadrature to take by comparing the contrast
             if take_fit_I:
-                ampl = abs(a.fit_res[0].params['period'].value)/2
-                print("taking I")
+                ampl = abs(a.fit_res[0].params['period'].value)/2.
+            elif (np.abs(max(a.fit_res[0].data) -
```
AdriaanRol: You are essentially taking the amplitude of the data here but taking it from the fit result. I would suggest changing this to directly taking it from the data (without the a.fit_res[0]), as this will only create confusion.  

I did not change this as I cannot test the qubit object but I would suggest we change this.

pycqed/instrument_drivers/meta_instrument/qubit_objects/qubit_object.py
```diff
@@ -421,26 +426,41 @@ def find_pulse_amplitude(self, amps,
             if n > max_n:
                 break
             else:
+                old_amp = ampl
```
AdriaanRol: Good renaming of variable, that should prevent some confusion.

pycqed/instrument_drivers/meta_instrument/qubit_objects/qubit_object.py
```diff
@@ -421,26 +426,41 @@ def find_pulse_amplitude(self, amps,
...
                 ampl_span = 0.3*ampl/n
```
AdriaanRol: Though I would also rename it in this line.

pycqed/instrument_drivers/meta_instrument/qubit_objects/qubit_object.py
```diff
@@ -421,26 +426,41 @@ def find_pulse_amplitude(self, amps,
...
+                if not min(amps) < ampl < max(amps):
+                    ampl_span *= 2
+                    amps = np.linspace(old_amp-ampl_span,
+                                       old_amp+ampl_span, 15)
+                    self.measure_rabi(amps, n=n, MC=MC, analyze=False)
+                    a = ma.Rabi_parabola_analysis(close_fig=close_fig)
+                    # Decide which quadrature to take by comparing the contrast
+                    if take_fit_I:
+                        ampl = a.fit_res[0].params['x0'].value
+                    elif (np.abs(max(a.fit_res[0].data) -
```
AdriaanRol: Same comment above applies here, try to refer to the data directly instead of through the fit_res object. 

I think that in the future I want to also make this find amplitude compatible when only a single quadrature is acquired (as is the case with optimal weights). But that is beyond the scope of this PR.

PR #200, opened 2017-04-19
Gauss Pulse Phase

lciorciaro:
Added possibility to define an arbitrary phase for a gauss pulse.


pycqed/instrument_drivers/meta_instrument/CBox_LookuptableManager.py
```diff
@@ -74,6 +75,11 @@ def __init__(self, name, **kw):
         self.add_parameter('Q_modulation', vals=vals.Numbers(), unit='Hz',
                            parameter_class=ManualParameter,
                            initial_value=20.0e6)
+        self.add_parameter('Q_Rphi', label='Phase of Rphi pulse',
```
NielsBultink: The name QR_phi seems a bit arbitrary to me. Looking at traditinoal polarcoordinates the phi is justiphied. 

Consistent with X180, Y180 would be Phi180. However, I see the need for an extra letter from the distinguashalble prameter point of view. Dont's have a better suggestion for now.

PR #201, opened 2017-04-19
Enh/#176/unit aware label

### LINKED ISSUE #176: Function to set SI_label for matplotlib
What it should do: 
take in a label and unit. Get the ticks of the specific axis, rescale them in a unit aware way and then set the label to label + unit with prefix.

AdriaanRol:
@fluthi @NielsBultink @elrama- 
A unit aware way of setting xlabel and ylabel, use this instead of the matplotlib defaul ax.set_xlabel/ax.set_ylabel 

The following snippet should serve as a minimal example of how it works. 

```python 
%matplotlib inline
import matplotlib.pyplot as plt
import numpy as np
from imp import reload
from pycqed.analysis.tools import plotting as pl


f, ax = plt.subplots()
x = np.linspace(-6, 6, 101)
y=np.cos(x)
ax.plot(x/1000, y*1e3)

pl.set_xlabel(ax, 'Time', 's')
pl.set_ylabel(ax, 'Distance', 'm')
```

And the resulting figure. 
screenshot 2017-04-19 22 59 03

This closes #176.

### DIFF
--- a/pycqed/analysis/tools/plotting.py
+++ b/pycqed/analysis/tools/plotting.py
@@ -3,7 +3,9 @@ Currently empty should contain the plotting tools portion of the
 analysis toolbox
 '''
 import numpy as np
-from pyqtgraph.units import SI_PREFIXES, UNITS as SI_UNITS
+
+SI_PREFIXES = 'yzafpnum kMGTPEZY'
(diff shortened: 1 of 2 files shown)


fluthi:
Yup, works nicely.

PR #210, opened 2017-04-22
Execfile snippet from python 2

AdriaanRol:
Convenience function. Added here so we don't have to define it in the init. 

@NielsBultink this is the snippet I showed you yesterday. If you approve you can merge it.


elrama-:
Teste and goes OK. Why is code the police complaining?

AdriaanRol:
@elrama- , code police is OK no?
Also if you are wondering, you can click the link and see what's wrong.

Issue #212, opened 2017-04-23
Better way to deal with calibration points/normalization and rotation of axes


AdriaanRol:
There is a whole range of analysis scripts for all kinds of time domain experiments. All of these have the central underlying assumption that data is acquired in two quadratures (I and Q) and contains calibration points. These points are then used to rotate and normalize the data. As we are moving towards multiplexed experiments where we use a single weight function per qubit this assumption breaks, especially when it is hard to easily add calibration points (as is currently the case for some QASM sequences). 

As we use inheritance for all these functions I think they should work with different types of input. I think the following should be supported (note that a checkbox means it already is supported). 
- [x] 2 quadratures with calibration points
- [ ] 2 quadratures with calibration values based on SSRO experiment 
- [x] single qaudrature with calibration points 
- [ ] single quadruature with calibration values based on SSRO experiment 
- [ ] precalibrated axis in single quadrature. 

It is important that a solution is found that in turn makes all Timedomain experiments work regardless of the fashion in which the data is acquired.

MiguelSMoreira:
Closed due to inactivity

Issue #178, opened 2017-04-11
Multiplexed SSRO analysis integration in PycQED


AdriaanRol:
There exists a two qubit SSRO analysis script by Niels in the PycQED scripts: `pycqed_scripts.personal.Niels.two_qubit_readout_analysis` . This script should be integrated in the measurement analysis of PycQED. I've copy pasted Niels' version here for reference. 

```python 
from pycqed.analysis import measurement_analysis as ma
import numpy as np
import pylab
# %matplotlib inline
from matplotlib import pyplot as plt
from numpy.linalg import inv

def two_qubit_ssro_fidelity(label, fig_format='png'):
    #extracting data sets
    # qubits=['q0','q1', 'q2', 'q3']
    states=['00','01','10','11']
    list_qubits=['q0', 'q1']
    nr_states = len(states)
    namespace = globals()

    #ma.SSRO_Analysis(label =label, auto=True, channels=['w0'])
    data=ma.MeasurementAnalysis(auto=False, label=label)

    #ma.SSRO_Analysis(auto=False)
    #extract fit parameters for q0
    w0_data=data.get_values('w0')
    w1_data=data.get_values('w1')
    lengths=[]
    i=0
    for nr_state, state in enumerate(states):
        if i==0:
            namespace['w0_data_r{}'.format(state)]=[]
            namespace['w1_data_r{}'.format(state)]=[]

    for nr_state, state in enumerate(states):
        namespace['w0_data_sub_{}'.format(state)]=w0_data[nr_state::nr_states]
        namespace['w1_data_sub_{}'.format(state)]=w1_data[nr_state::nr_states]
        lengths.append(len(w0_data[nr_state::nr_states]))

    #capping off the maximum lengths
    min_len = np.min(lengths)
    for nr_state, state in enumerate(states):
        namespace['w0_data_sub_{}'.format(state)]= namespace['w0_data_sub_{}'.format(state)][0:min_len]
        namespace['w1_data_sub_{}'.format(state)]= namespace['w1_data_sub_{}'.format(state)][0:min_len]
    for nr_state, state in enumerate(states):
        namespace['w0_data_r{}'.format(state)]+=list(namespace['w0_data_sub_{}'.format(state)])
        namespace['w1_data_r{}'.format(state)]+=list(namespace['w1_data_sub_{}'.format(state)])
    # for state in states:
[...]

PR #214, opened 2017-04-24
Bug/#213/uhfqc detector normalization

AdriaanRol:
@NielsBultink I have updated the detectors for the UFHQC to support the different acquisition modes. 

Can I ask you to look at the changes in the detector functions file in detail? Also please don't merge if OK as this still needs to be tested (both on your setup (tek_transmon)  and mine (CC_transmon). 

Here is a summary of the changes. 
- new result_logging parameter as part of the UFHQC detectors that supersedes the crosstalk suppression parameter. See summary below. 
- Added docstrings and comments that should clarify some things 
- updated the relevant files (qubit objects multi qubit module) to work with this update 
- added the multiplexed RO analysis file based on the code by Niels (should help addressing #178 ) 
- DDM if statement was defined twice in tek transmon object, so I removed one. 
- I removed the rotate and normalize option from the UHFQC detector as this is replaced with the optimal weight normalized detector. 

## Result logging parameter change
The detectors now have a parameter "result_logging_mode" that has three options. 
Below is an excerpt from the docstring: 
```
        result_logging_mode (str) :  options are
            - raw        -> returns raw data in V
            - normalized -> returns normalized data using the crosstalk
                            suppression matrix and offset defined in the
                            UHFQC instrument. Requires optimal weights.
            - digitized  -> returns fraction of shots based on the threshold
                            defined in the UFHQC. Requires optimal weights.
```

If one wants to calibrate the SSRO threshold and inversion matrix, the result_logging_mode should be set to normalized with the correction matrix set to the Identity matrix. 

I have


examples/qubit_tuneup/CPhase_tuneup.py
```diff
@@ -31,7 +31,7 @@
               DataT.RO_acq_weight_function_I()],
     nr_averages=AncT.RO_acq_averages(),
     integration_length=AncT.RO_acq_integration_length(),
-    cross_talk_suppression=True)
+    result_logging_mode='normalized')
```
NielsBultink: as discussed with @AdriaanRol: change 'normalized' to 'lintrans' as it is a general linear transformation, not just normalization.

AdriaanRol:
@NielsBultink , Renamed "normalized" mode to "lin_trans" as discussed. Includes update to docstring.

pycqed/instrument_drivers/meta_instrument/qubit_objects/qubit_object.py
```diff
@@ -103,7 +103,8 @@ def measure_spectroscopy(self):
     def measure_transients(self):
         raise NotImplementedError()
 
-    def calibrate_optimal_weights(self):
+    def calibrate_optimal_weights(self, MC=None, verify=True,
```
NielsBultink: I would prefer to have a more general 'set_weights( )' function. Based on three qubit object parameters(all booleans) a different calibration/setting is performed:
- SSB: determins whether we use single sideband or double sideband measurement
- optimized weights: determines whether weight functions are measured or standard weight functions are set 
- one weight functions, only loads weight functions into the I channel

AdriaanRol:
@NielsBultink small status update. 
I have tested the changes to the detector on the CC_transmon. This is now working. 
Before we can merge this, several things need to be done. 
- [x] Test on CBox/CC driven transmon 
- [ ] Test on tektronix driven transmon (relates to merging your branch)
- [ ] cleanup new correlations detector to fit in the same paradigm 
- [ ] resolve #212 as using only one weight doesn't play well with analysis. 

Let me know if I missed something (and if this is OK with what you had in mind based on our discussion earlier).

AdriaanRol:
@NielsBultink I would like to merge this as it's been in the backlog for quite a while. I think we've discussed most of the things in here. 
Further updates are on the Starmon branch but I would like to have that in a separate review.

PR #222, opened 2017-05-04
Added optional argument of "datadir" folder to analysis functions

GitJaap:
Very small non-intrusive changes.
Added optional argument of "datadir" folder to analysis functions and hdf5 saving so loading the config file is no longer required when just using some timestamp related functions.

### DIFF
--- a/pycqed/analysis/measurement_analysis.py
+++ b/pycqed/analysis/measurement_analysis.py
@@ -30,7 +30,7 @@ from pycqed.analysis import composite_analysis as ca
 
 try:
     import qutip as qtp
-    from pycqed.analysis.tomography import Tomo_Multiplexed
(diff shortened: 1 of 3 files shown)


pycqed/analysis/measurement_analysis.py
```diff
@@ -30,7 +30,7 @@
 
 try:
     import qutip as qtp
-    from pycqed.analysis.tomography import Tomo_Multiplexed
+    #from pycqed.analysis.tomography import Tomo_Multiplexed
```
GitJaap: sorry for this change, is not related to the pull request, but why is the line here anyway? I would expect the tomo to be imported separately by a user

elrama-: You are right! See PR #223 (Livio fixed this)

PR #228, opened 2017-05-19
Increased resolution in HDF5 data

quantumCryoCoder:
Increase the resolution of the data saved in the HDF5 files.

Data with high resolution were rounded before saving


elrama-:
There is an easy test we could implement:
1. set up a fake test
2. measure two points, which if not saved with enough precission will be stored as the same value.
3. assert that the saved values are different

I will try to implement this next week

AdriaanRol:
@elrama- I'll assign you to this PR and mark the request changes button for the test you propose.

PR #231, opened 2017-05-19
Fixed problems with 'resonator fit'

quantumCryoCoder:
Minor fix: the abs value of Qi is implemented in the fit function. This minor fit is necessary because of negative values returned in same cases.

Major fix: the "complex" option of the fit was not working because the 'params' were not prepared.


fluthi:
Looks good to me!

AdriaanRol:
has not been merged due to conflicts, by now super-ceded by #341

PR #232, opened 2017-05-19
Fix for MC instr params in Rabi method

quantumCryoCoder:


elrama-:
I helped @quantumCryoCoder do the fix, so I will explain:

After the PR that implemented instrument parameters, there are still some places where the qubit object does actions like self.MC.set_sweepfunction() instead of self.MC().set_sweepfunction()

In this case, the calibration of the rabi amplitude in the parent class of all qubits (qubit_object) was fixed

PR #234, opened 2017-05-19
Move_to_dac_value function in qubit object

quantumCryoCoder:
Qubit.move_to_dac_value(val)
sets the IVVI channel linked to the qubit to the 'val' specified.
Moreover, it updates the dac value parameter in the qubit object

### DIFF
--- a/pycqed/analysis/measurement_analysis.py
+++ b/pycqed/analysis/measurement_analysis.py
@@ -3838,7 +3838,7 @@ class Homodyne_Analysis(MeasurementAnalysis):
         else:
             ax.plot(self.sweep_points, fit_res.best_fit, 'r-')
             f0 = self.fit_results.values['f0']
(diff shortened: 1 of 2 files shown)


elrama-:
seems okay. if it works i think we can merge.
@AdriaanRol does it make sense to have a test? it is a wrapper of the ivvi call,,

AdriaanRol:
@elrama- If you approve the review then you can merge :)

elrama-:
@fluthi I think the two changes you required belong to the IVVI driver rather than to the qubit object right? (This would mean, another pull-request with those changes)

If I don't remember it wrongly, in old PycQED the speed and step of the channels was set in the IVVI instrument, not in the qubit object.

AdriaanRol:
Rejecting this PR has there are conflicts and requested changes and no one seems to be working on it.

PR #235, opened 2017-05-21
Enh/fridge mon

### LINKED ISSUE #227: Improvements to fridge monitor
- [x] Also extract timestamp of last update 
- [x] If fridge is no longer updating, give warnings 

- [x] Add monitoring of presures

AdriaanRol:
Updates to the fridge monitors in PycQED. This closes #227 

Changes include 
- now also displays pressures 
- includes the last update time 
- unit aware temperatures  and pressures 
- gives a warning if the website is no longer updating 
- separated the settings of what is extracted from the driver itself 
- included Kelvin and Bar in the units that are compatible with unit aware rescaling. 
- Updated the example notebook 

screenshot 2017-05-21 18 13 50


elrama-:
Seems okay, I am going to pull it in la vespa next time to try it. Two minor questions:
1. Why is it high priority? was it not working before?
2. Are there tests?

AdriaanRol:
@elrama- 
1. It is high priority because of the problems with the fridge last week. The fridge monitor before was working but it was not possible to see if the website was not updating. The fridge monitor now also extracts the time of the last update from the website and raises a warning if this is older than 6 mins. 

2. No there are no tests and there also will not be tests. This is because the tests would have to rely on a specific fridge monitor being up, and even then cannot be very specific as we cannot test if the right numbers are extracted (these change). This could be worked around by a static website that is always on but I think that is more trouble than it is worth. 

Let me know if this is OK

PR #236, opened 2017-05-22
Measurement Analysis - hanger fit and spectrocopy

quantumCryoCoder:
Two improvements implemented in the measurement_analysis file:
- one big fix in the "Qubit_spectroscopy_Analysis" class
- one small cosmetic improvement in the "fitting resonator" plot

Big fix
- The problem:
Currently, after a spectroscopy measurements, data are converted in "distance" form the ground state. The function a_tools.calculate_distance_ground_state() is called with both I/Q values or amplitude/phase values. Once amplitude/phase values are passed most of the information is lost, resulting in a flat and noisy trace. As a consequence the spectroscopic peak is not found.
- The change:
If amplitude and phase are measured, the spectroscopic peak is found using only the amplitude data. This has been tested and working properly

Small fix
- The problem:
Due to a wrong conversion Hz/GHz the plot generated with "homodyne analysis" was not readable.


fluthi:
Addressing the big fix:
I don't think this is the best way to go because there are situations where most of your signal is going to be in the phase. Hence, in these cases, you will not get a reliable frequency either.
What might be a good solution is to check all three signals (Amplitude, Phase and Distance) to see which one gives the most reliable peak and choose that one.

AdriaanRol:
@quantumCryoCoder I've been using this and am quite happy with it (it works). I do agree with the issue raised by @fluthi but for now I would be OK with merging this. 

I will not approve this PR as I will leave approving or requesting changes to Flo for the moment.

AdriaanRol:
This has been super-ceded by #341

Issue #189, opened 2017-04-12
Rescaling tomography to ensure proper noise weighting in linear inversion  


AdriaanRol:
Today we did some testing with @lciorciaro , @elrama- and @leodicarlo and found some issues to happen when noise in the correlations are not weighted properly.

We tested that rescaling fixes this problem and have a working version now. We need to properly implement these changes and update the Tomo. 

@GitJaap you might also be interested in this.

MiguelSMoreira:
Closed due to inactivity

PR #237, opened 2017-05-23
Bell Tomo Analysis

lciorciaro:
This is a pull request related to #189. It is not ready to be merged into master, I am putting this out here to start some discussion and work on it.

The only real change I have made until now is making the `Tomo_Multiplexed` class inherit from `MeasurementAnalysis` like it is supposed to be, and added saving of the MLE analysis data to the data file.

What still needs to be done is cleaning up the actual analysis code itself, including the fix proposed in #189, i.e. weighting the errors properly. The code in this branch currently contains the changes that were made during a discussion with Ramiro, Adriaan, and Leo.

The code on lines 668-769 is currently very messy, and I am not sure it is completely correct. @elrama- @AdriaanRol, I would be happy to rewrite it in a nice and correct way, but I need some discussion on what is the correct way to do it first. I know it is not top priority now, so let me know if you have time and would like to help me get started.


GitJaap:
Hi,

As we discussed. I think the renormalization and scaling by noise should happen outside the tomo toolbox. This is due to the fact that the toolbox is initialized with already averaged data(the implementation of which  depends heavily on your measurement setup(how many calibration points etc). ).

So all you have to do is tidy up the scaling in the measurement_analysis extending class but you should be able to leave the tomo toolbox as it is. 

One more thing. Maybe you can update the tomo with the latest version from pycqed 2, because it includes printing of the labels and rotation matrix order  so you can check if that order actually corresponds with your setup.

AdriaanRol:
@lciorciaro given the recent fixes and the age of this PR. Would it make sense to reject/close this PR?

PR #238, opened 2017-05-30
Complex Cosine fitting

quantumCryoCoder:
Implements a joint cosine fit on both I and Q quadrature.

- Before the change:
separate fits were performed on I and Q quadrature
This approach returned errors or wrong fit results with signal mainly in one quadrature

- After the change:
Cosine fit of I and Q quadrature is done by considering both at the same time. Only one frequency is used.
The fit is robust (tested) with information mainly on one quadrature or on both quadratures

- Temporary flag
The complex fit is called with the flag fitting_model='complex'. For example:

ma.Rabi_Analysis(timestamp=xxx, fitting_model='complex')

A fit with the previous routine is done if the fitting_model is not specified (back compatibility)

I suggest to use this flag for a probation period. If we are happy with the new routine we can make the 'complex' fitting the default one.

### DIFF
--- a/pycqed/analysis/fitting_models.py
+++ b/pycqed/analysis/fitting_models.py
@@ -204,6 +204,33 @@ def HangerFuncComplex(f, pars):
     return S21
 
 
+def CosComplex(t, pars):
+    '''
+    Cos function in a 2D plan.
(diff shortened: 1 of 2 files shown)


AdriaanRol:
Closing this PR as it does not seem to be picked up by anyone.

PR #239, opened 2017-05-30
Added a new span utility as requested in the code meeting

AdriaanRol:
- Added two functions for creating a span to utilities.general
- Added a gen_sweep_pts function that accepts different kinds of input for generating arrays
- added tests for the functions 
Let me know if this is what you are looking for. 

N.B. I am on purpose not giving an example on how to use it here, I would like to know if you can see it from the code review :)


AdriaanRol:
@quantumCryoCoder 
1. I like it as independent functions, I think it can provide some utility to have this. 
2. num is choses as the variable name as this is the same variable name numpy uses, I don't want to introduce yet another convention. 
3. Done, I updated the docstrings and comments, well spotted. 

Let me know if this addresses your comments.

AdriaanRol:
@fluthi well spotted. I can't believed I missed that. Thanks

AdriaanRol:
@fluthi I fixed your bug. Thanks

AdriaanRol:
@nathanlangford 
- I've renamed the function as you suggested. 
- I agree that the trick is inelegant but sadly the numpy arange does not natively support endpoint as an option. 
- I agree on the weaknesses you describe, I think it is a bit out of scope for this PR to address it here. A better solution (that would work for any function we make) would be to support it in the level of the measurement control itself. I need to think a bit more on this but it is definitely possible. 

Could you let me know if this addresses your issues (i.e., is this OK to merge?)

nathanlangford:
Don't get me wrong, I like the endpoint trick, I just think it's inelegant
to put it in two different places in the code.  You could fix this by
writing a small wrapper for np.arange, but I'm not wedded to this idea.

Re the weakness for the span, I would at the very least allow a default
value for center being 0, so that it is more flexible.  This would just
require creating a nested loop, rather than a loop with an "and" in the
condition.

Cheers,
Nathan.

AdriaanRol:
@nathanlangford I've added center=0 as a default for the gen_sweep_points function. 
I didn't do it for the underlying functions as it would require me to switch around the order of arguments (which I think is bad) or also give the other arguments default values. 

Furthermore I've added type hints (python3.5 feature) I think that is nice. 

Let me know if this addresses your issues.

AdriaanRol:
@fluthi , @quantumCryoCoder could you hit the approve button if my changes address your issues ( I think they do) it is currently preventing me from merging this.

PR #246, opened 2017-06-14
Qasm compiler dev

AdriaanRol:
Fixes #issuenumber.

- Support for long wait operations has been added to the Qumis assembler, this feature resolves a bug in the example qasm file. 
- Bug relating to wrong config has been fixed 
- basic tests added for the compiler that serve as an example, it tests for valid code according to the Assembler. 

@gtaifu let me know what you think of this.


AdriaanRol:
Added support for map in the qubit config (closes #247 )

AdriaanRol:
@gtaifu , Could you review this? I'm ready to send the next PR (that solves most of the bugs I've reported) but would like to keep the changes to be reviewed managable.

PR #261, opened 2017-06-18
Bug/butterfly

### LINKED ISSUE #257: Butterfly is broken with CC transmon
### Steps to reproduce
1. Measure a butterfly using the CC transmon 
2. Encounter a linalg error 

Blocking for any kind of restless experiment. Debugging now.

### System
**Proj/starmon**

**4e7e3813c0**

AdriaanRol:
Fixes #257.

- bug in discrimination fidelity analysis solved (was related to auto range of the plots) 
- fixes inconsistency between discrimination fidelity as defined in SSRO analysis and as defined in single quadrature discrimination fidelity analysis (factor 2 related to conservative and current definition) 
- tests added for single quadrature SSRO discrimination 
- better figures for single qaudrature discrimination analysis 
    - single quadrature analysis shows both components of the double gaussian fit 
    - 2D histograms now unit aware 
- made SSRO analysis work better with single quadratrue data 

Figures from the test datasets 

2d-histograms_rot_0 0 deg

shots_x_vs_y


AdriaanRol:
@lciorciaro 
1: I have tested this on our setup and as far as I know that bug is still there. For now this solves an existing bug and as such I think this should be merged. The other one may or may not be related but I think that that is unrelated to the quality and scope of this PR. 
2: I removed the Ram_Z analysis from this PR.

PR #269, opened 2017-06-21
Multiple AWG support

### LINKED ISSUE #120: Support for multiple AWGs
The current system has the `station.pulsar` object hardcoded to the sweep functions and there is only a single AWG in the `Pulsar` class. This makes it impossible for us to use multiple AWG-s with the current code (as far as I can tell). However we need multiple AWGs to do experiments with multiple qubits. Is there a simple workaround to this? One way to fix this I could imagine is specifying the AWG name in addition to the channel in the pulse parameters and make the `Pulsar` class take this into account.

antsr:
This is a rather big pull request that fixes #120 by adding support for multiple AWG-s. 

The main changes proposed in this pull request:
- Support defining channels for pulsar on different physical AWG-s by adding the argument `AWG` to `Pulsar.define_channel`. Default value for the AWG name is taken from `self.AWG.name` so old code should not break if `pulsar.AWG` was defined.
- Add an instrument parameter `master_AWG` which should point to the instrument that triggers all the other (slave) AWGs. When starting the AWGs with `Pulsar.start()` the master AWG is started last, after all the slave AWGs are waiting for a trigger.
- Promote Pulsar to a qcodes `Instrument` as it has 2 instrument parameters now (default and master AWG)
- Add a parameter `'slave_AWG_trig_channels'` to the `station.sequence_config` which is a list of channel names that are used to trigger the slave AWGs.
- Until now the pulsar channel options (high, low, etc.) were updated to the values defined in the AWG object before programming the AWG. Having the same parameters defined in multiple places created confusion. This pull request changes the code so that `Pulsar.program_awgs` automatically updates the AWG parameters to the values defined in pulsar. This is compatible with the view that `Pulsar` is an abstraction of all the physical AWGs, which themselves should not be exposed much in the codebase.
- Because different AWGs can have different sampling rates, many of the functions in `Element` now depend on the channel. Particularly, the `tvals` returned by `Element.(ideal_/normalized_)waveforms()` is now a dictionary with `tvals` for all the channels.
- `Element` does not store a private copy of the channel parameters now, but uses the ones defined in `Pulsar`. Because of this, a pulsar object needs to be passed to `Element` on initialization.
- Add a virtual Tektronix AWG5014 class to help testing and add a test for programming multiple AWGs.
- several other small fixes

Could you please review this and test that it does not break things for you @AdriaanRol and others?

One problem with the current code is that to reliably avoid pulses starting at negative times (and thus producing an error) because of the rather high pulse delays related to triggering the different instruments, one needs to set the readout fixpoint to a value higher than the length of the pulse sequence plus the slave AWG delays.


AdriaanRol:
@antsr , great to see this PR! I think this is a feature we definitely want to have. I don't have time to look at this immediately but I will give it a detailed review and will probably leave some questions. 

I'm currently not using an AWG5014 in my setup so I cannot test this myself but I will try to find someone who is in order to test this.

AdriaanRol:
Hi @antsr , I don't know if you noticed but your test build is failing.

antsr:
Thanks for pointing out, will fix shortly.

AdriaanRol:
@antsr , I have discussed with @NielsBultink and he has agreed to test your AWG changes when he has finished taking final data for his paper, the current estimate for that would be around the middle of next week.

NielsBultink:
Checked out the branch after pulling in the master. Could do resonator spectroscopy and basic time domain measurements on functionality. However, no device was connected. Approving the pull request based on this test.

PR #270, opened 2017-06-22
Added error check whenever QWG.start is called + fix for bad validator 

AdriaanRol:


AdriaanRol:
@wvlothuizen I've addressed your changes. Can you take a look.

PR #276, opened 2017-06-27
fix: DSB demodulation scaling with UHFLI

### LINKED ISSUE #275: UHFLI detector scaling
### Steps to reproduce
1. Measure heterodyne spectroscopy with UHFLI, single_sideband_demodulation = True
2. Measure heterodyne spectroscopy with UHFLI, single_sideband_demodulation = False
3. See that the amplitudes differ by a factor of 2.

### Expected behaviour
Regardless of the SSB demodulation setting, the result should be scaled correctly to the amplitude of the measured pulse.

### Actual behaviour
Using double sideband demodulation the returned value is half the amplitude.

### System
**operating system**
Windows 7 64 bit

**PycQED branch**
Proj/Bluefors1 (updated with master)

**PycQED commit**
24841015b457ba38f70ad69752f99cc1577dcf7b

### Labels to be added
type:bug:
prio:medium:
cat:UHFLI:

antsr:
Fixes #275 

- Change the rotation matrix weights to correctly scale UHFQC output when using double sideband demodulation

@AdriaanRol

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
@@ -627,6 +627,7 @@ class UHFQC(Instrument):
                  np.array(cosI))
         self.set('quex_wint_weights_{}_real'.format(weight_function_Q),
                  np.array(sinI))
+		# the factor 2 is needed so that scaling matches SSB downconversion
         self.set('quex_rot_{}_real'.format(weight_function_I), 2.0)
         self.set('quex_rot_{}_imag'.format(weight_function_I), 0.0)
         self.set('quex_rot_{}_real'.format(weight_function_Q), 2.0)
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
@@ -631,9 +631,9 @@ class UHFQC(Instrument):
             weight_function_Q))
         eval('self.quex_wint_weights_{}_imag(np.array(cosI))'.format(
             weight_function_Q))
-        eval('self.quex_rot_{}_real(1.0)'.format(weight_function_I))
+        eval('self.quex_rot_{}_real(2.0)'.format(weight_function_I))
         eval('self.quex_rot_{}_imag(0.0)'.format(weight_function_I))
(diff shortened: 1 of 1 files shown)


antsr:
I have not yet tested the fix on physical hardware, but it should work.

AdriaanRol:
@antsr , Could you add a comment as to why the factor 2 is needed?

antsr:
The integration weights set up by the UHFQC.prepare_SSB_weight_and_rotation() calculate the integrated result as:
```
out1 = ∑ᵢ ( in1ᵢ cos(ω i / fs) + in2ᵢ sin(ω i / fs) )
out2 = ∑ᵢ ( in1ᵢ sin(ω i / fs) - in2ᵢ cos(ω i / fs) )
```
or in complex form
```
out1 + j out2 = ∑ᵢ (in1ᵢ - j in2ᵢ) exp(j ω i / fs)
```
`ω` is the intermediate angular frequency and `fs` is the sampling frequency, `in1ᵢ` and `in2ᵢ` are the arrays of samples acquired by the UHFQC in channels 1 and 2 and `out1` and `out2` are the weighted integration results.

The formula for UHFQC.prepare_DSB_weight_and_rotation() in the old version is the same, except that the weights corresponding to input channel 2 are zero:
```
out1 + j out2 = ∑ᵢ in1ᵢ exp(j ω i / fs)
```

Now if we apply an ideal signal at the input of the UHFLI
```
in1ᵢ + j in2ᵢ = A exp(j ω i / fs + j ϕ)
```
the output for the SSB case is
```
out1 + j out2 = ∑ᵢ A exp(-j ω i / fs - j ϕ) exp(j ω i / fs) = A N exp(-j ϕ)
```
and in the DSB case:
```
out1 + j out2 = ∑ᵢ A cos(j ω i / fs + j ϕ) exp(j ω i / fs) = 
              = A/2 ∑ᵢ (exp(2 j ω i / fs + j ϕ) + exp(-j ϕ)) = 
              = A N exp(-j ϕ) / 2 + A exp(j ϕ) (1 - exp(2 j ω N / fs)) / (2 - 2 exp(2 j ω / fs))
```
It is clear that in the DSB case the absolute value of `out1 + j out2` is half of that of the SSB case in the big `N` (number of integration samples) limit.

To fix this I doubled the rotation matrix elements (with what the integration results for the two input channels are multiplied) for the DSB case.

antsr:
Also tested on real hardware now and working as expected.

PR #277, opened 2017-06-27
Enh/demonstation helpers

stanvn:

Fixes something in the QC_demonstrator repository.

- QX_detecotor for multiple qubits 
- Fixed bug in the QASM_loader, some how the QASM_loader removed the last character of the last line.
- Separated  the different helper function in different  file, so each helper function can be imported Separately. 
- Added Tess Connect to the helper function for automatic connection to tess  

@AdriaanRol


stanvn:
Removed the tess live plot parameter, because it is not doing anything right now. I'll put it back in place when I continue with the live plotting.

Issue #208, opened 2017-04-21
Integrate new unit aware xlabel and ylabel in measurement analysis


AdriaanRol:
#201 added the new set_xlabel and set_ylabel feature. This should be used in the default analysis scripts (which is also in need of a bit of a cleanup).

PR #281, opened 2017-07-09
Update to use unit aware labels in standard meas analysis

AdriaanRol:
Relates to #208

- make the standard measurement analysis use the unit aware plotting by default


brianzi:
This looks good. I have one question @AdriaanRol: Is there an easy way to tap into the unit handling, to fix cases where the stored units are not SI base units? It's just cosmetics, but having kilomillivolts in the default plots kind of means I cannot show them to anyone :)

AdriaanRol:
@brianzi we have already discussed this offline but for the sake of record keeping, the unit aware labels from this PR only get applied when the unit is a known SI unit. The bug you are talking about is part of the default pyqtgraph plotting library. 

A PR to that repository would solve this issue.

PR #291, opened 2017-07-23
Enh/data saving

### LINKED ISSUE #138: Not all parameters reload when using gen.load_settings_onto_instrument
While reloading the qubit object parameters, the dac_channel parameter does not get reloaded.

Don't understand exactly why

AdriaanRol:
Fixes #138 .

- new hdf5 datasaving and loading functions that allow storing arbitrary nested dictionaries
- new load_settings_onto_instrument_v2 that loads using the new dataformat 
- new snapshot saving added to the MC 
- small tweaks to the load_settings_onto_instrument_v2 version (no longer tries to set values that are not settable). 

- added tests for all functionality 
- tested on starmon setup 

@elrama- , I think you originally reported this. This should solve that problem and allow saving config dictionaries as well as integration weight functions to your datafile. 
@brianzi , we have discussed the data saving in some detail. Wondering what you think on how it turned out.


pycqed/tests/test_hdf5_datasaving_loading.py
```diff
@@ -0,0 +1,175 @@
...
+from qcodes import station
+from pycqed.analysis import analysis_toolbox as a_tools
+
+
+class Test_HDF5(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(self):
+        self.station = station.Station()
+        # set up a pulsar with some mock settings for the element
```
brianzi: A pulsar?

AdriaanRol:
@brianzi uh oh ... 

Removed the pulsar comment from all tests where it doesn't belong.

pycqed/measurement/hdf5_data.py
```diff
@@ -124,9 +125,135 @@ def encode_to_utf8(s):
...
     return s
+
+
+def write_dict_to_hdf5(data_dict: dict, entry_point):
+    """
+    Args:
+        data_dict (dict): dictionary to write to hdf5 file
+        entry_point (hdf5 group.file) : location in the nested hdf5 structure
+            where to write to.
+    """
```
GitJaap: What happens if you enter a list? Then you need to make the list into a dict first, so you could also try to catch that case here first and then when loading convert it back to a list.

btw nice tip on converting list to dict: dict_from_list =dict(zip(range(len(dic)),dic))

pycqed/measurement/hdf5_data.py
```diff
@@ -124,9 +125,135 @@ def encode_to_utf8(s):
...
+            # I create special string, note that this can create
+            # unexpected behaviour if someone saves a string with this name
+            entry_point.attrs[key] = 'NoneType:__None__'
+
+        elif isinstance(item, dict):
+            entry_point.create_group(key)
+            write_dict_to_hdf5(data_dict=item,
+                               entry_point=entry_point[key])
+        elif isinstance(item, list):
+            if len(item) > 0:
```
GitJaap: elif isinstance(item, list):
                if isinstance(item[0], (np.int64, np.float64, str, np.float, float, np.float32,int)):
                    item = np.array(item)
                    h5file[path + key] = item
                else:
                    dict_from_list =dict(zip(range(len(item)),item))
                    dict_from_list['_h5save_type'] = 'list'
                    recursively_save_dict_contents_to_group(h5file, path + key + '/', dict_from_list)

this is just my snippet of code on making a list where recursively_save_dict_contents_to_group is the same as your function

pycqed/measurement/hdf5_data.py
```diff
@@ -124,9 +125,135 @@ def encode_to_utf8(s):
...
+                    group_attrs['list_type'] = 'generic_list'
+                    group_attrs['list_length'] = len(item)
+                    write_dict_to_hdf5(
+                        data_dict=list_dct,
+                        entry_point=entry_point[key])
+            else:
+                # as h5py does not support saving None as attribute
+                entry_point.attrs[key] = 'NoneType:__emptylist__'
+
+        else:
```
GitJaap: don't forget to check tuples:
elif isinstance(item, tuple):
//now you can do the same as with lists

AdriaanRol: Will do

pycqed/measurement/hdf5_data.py
AdriaanRol: I'll ad a test and see what happens (and fix it if it breaks) but I think it should work ...

pycqed/measurement/hdf5_data.py
AdriaanRol: Indeed the same snippet :)

pycqed/measurement/hdf5_data.py
AdriaanRol: @GitJaap , Now also supports generic tuples.

AdriaanRol:
@brianzi @GitJaap , Could either one of you approve this PR? I have addressed your comments.

pycqed/measurement/hdf5_data.py
```diff
@@ -124,9 +125,150 @@ def encode_to_utf8(s):
...
+            logging.warning(
+                'Type "{}" for "{}" (key): "{}" (item) at location {} '
+                'not supported, '
+                'storing as string'.format(type(item), key, item,
+                                           entry_point))
+            entry_point.attrs[key] = str(item)
+
+
+def read_dict_from_hdf5(data_dict: dict, h5_group):
```
lciorciaro: Why is the first argument needed? Wouldn't it be better to create a new dictionary in this function and then just return it. Then the user calls

`loaded_dict = read_dict_from_hdf5(my_h5_group)`

instead of

`loaded_dict = {}
read_dict_from_hdf5(loaded_dict, my_h5_group)`

AdriaanRol: @lciorciaro , the first argument exists to allow nested calls of this function all adding things to the same dictionary. I've found this to be a very convenient and clear.

pycqed/measurement/hdf5_data.py
```diff
@@ -184,14 +190,18 @@ def write_dict_to_hdf5(data_dict: dict, entry_point):
...
                                 elt_type, key, item))
                         entry_point.attrs[key] = str(item)
-                # Storing of generic lists
+                # Storing of generic lists/tuples
                 else:
                     entry_point.create_group(key)
                     # N.B. item is of type list
                     list_dct = {'list_idx_{}'.format(idx): entry for
                                 idx, entry in enumerate(item)}
+                    print(list_dct)
```
GitJaap: ^^

PR #292, opened 2017-07-25
Enh/uhfqc statistics detectors

AdriaanRol:
Detectors and updates to the UHFQC for the state pattern mode

@nielshaandbaek , no need to respond quickly, I'm mostly tagging you as this relates the UFHQC drivers

### DIFF
--- a/pycqed/measurement/detector_functions.py
+++ b/pycqed/measurement/detector_functions.py
@@ -134,8 +134,8 @@ class QX_Hard_Detector(Hard_Detector):
         self.set_kw()
         self.detector_control = 'hard'
         self.name = 'QX_Hard_Detector_Fast'
-        self.value_names = ['F']
-        self.value_units = ['|1>']
+        self.value_names = []
+        self.value_units = []
         self.times_called = 0
         self.__qxc = qxc
         self.num_avg = num_avg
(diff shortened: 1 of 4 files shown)


pycqed/measurement/detector_functions.py
```diff
@@ -1524,19 +1572,33 @@ def define_correlation_channels(self):
...
+                    # correlation mode is turned on in the
+                    # set_up_correlation_weights method
                     break
             if correlation_channel < 0:
                 raise ValueError('No free channel available for correlation.')
 
             self.correlation_channels += [correlation_channel]
 
     def set_up_correlation_weights(self):
-        # FIXME: add option to use thresholding (quex_rl_soure(5))
```
lciorciaro: It looks like this has been fixed, so this line can be removed

pycqed/measurement/detector_functions.py
```diff
@@ -2026,3 +2296,20 @@ def get_values(self):
     def finish(self):
         if self.AWG is not None:
             self.AWG.stop()
+
+
+class RTO1024_detector(Hard_Detector):
```
lciorciaro: This class should not go in this pull request, as it uses the scope driver that is under development, has not been tested yet, and has nothing to do with the UHFQC detectors.

PR #293, opened 2017-07-25
Awg8 driver

AdriaanRol:
@antsr , @NielsBultink , @nielshaandbaek 
Initial driver for the HDAWG8 

This PR contains 
- A driver for the AWG-8 
- the ZI_shell for the AWG8 for the AWG8 (used to as a basis for the instrument driver)
- a base class that is intended as a general base class for all future ZI instruments (UHFQC/AWG8 and future devices) 
- an example notebook where two example waveforms can be loaded onto the AWG8. 

There are several problems with the driver that I do not intend to address in this PR, but do need to be addressed to make this a driver that up to standards and usable by people not intimately familar with the AWG8. Some of these issues also relate to the UHFQC. 

The list of issues: 
- There are no docstrings/documentation for the specific nodes. A per node docstring would help a lot in the daily use when someone does not know what a certain node does. 
- There are *no units* specified for values. e.g., I can set amplitudes between 0 and 1 but I do not know if these are in units of V or in arbitrary units relating to what the output mode is set to. 
- Most things are either floats or ints. It would be good to expose a more friendly value to the user, to give an example. The parameter 'raw_sigouts_1_mode' can be either 0 or 1. One of the two corresponds to "direct mode" and the other to "amplified mode". It would be good to replace this with an Enum('direct', 'amplified'). 
- There is a general lack of documentation beyond the initial information compilation. 

Some functions do not work. 
- the restart_device() method of ZIShell does not work 
- the node 'raw/sigouts/*/mode' cannot be found using the find command (that should discover nodes) but can be set by hand. Now it is explicitly added by hand. 
- the sigouts/*/offset parameter goes between 0 and 2.24e-44, It is not clear what the significance of these numbers is but they do clearly control the offset on the AWG 8.


pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_base_instrument.py
```diff
@@ -0,0 +1,227 @@
...
+        "{par_type}_node_pars_{dev_type}.json"
+
+        '''
+        # set the file names to write to
+        dev_type = self._dev.daq.getByte(
+            '/{}/features/devtype'.format(self._devname))
+        # Watch out this writes the file to the directory from which you
+        # call this command
+        s_file_name = 's_node_pars_{}.json'.format(dev_type)
+        d_file_name = 'd_node_pars_{}.json'.format(dev_type)
```
antsr: As the comment says, the files will not be written to the zi_parameter_files folder as is specified by the docstring. Should be changed similar to the current UHFQC driver.
```
path = os.path.abspath(__file__)
dir_path = os.path.dirname(path)
self._s_file_name = os.path.join(
    dir_path, 'zi_parameter_files', 's_node_pars_{}.json'.format(dev_type))
self._d_file_name = os.path.join(
    dir_path, 'zi_parameter_files', 'd_node_pars_{}.json'.format(dev_type))
```

pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_HDAWG8.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+from . import zishell_NH as zs
+from qcodes.utils import validators as vals
+from .ZI_base_instrument import ZI_base_instrument
+from qcodes.instrument.parameter import ManualParameter
+
+
+class ZI_HDAWG8(ZI_base_instrument):
+    """
+    This is PycQED/QCoDeS driver driver for the ZI_HDAWG8.
```
antsr: -> This is PycQED/QCoDeS driver for the Zurich Instruments AWG-8.

NielsBultink: Good point. In my opinion we should change the name to HDAWG. It is now a merge between the code name AWG8 and the name featured on the box. HD stands for High definition I suppose.

PR #296, opened 2017-07-27
Enh/spec analysis

tstavenga:

Fixes #issuenumber. does not fix any current Issue

- Added fitting to the base class
- Added Spectroscopy and resonator class for all general spectroscopy measurements
- Added Nanowire class for Nanowire specific functions
- Updated fitting models to have a guess function for the resonator fitting

@AdriaanRol @fluthi  

- travis test suite passes

- follow the PEP8 style guide 
- have tests for the code

### DIFF
--- a/pycqed/analysis/fitting_models.py
+++ b/pycqed/analysis/fitting_models.py
@@ -237,6 +237,62 @@ def SlopedHangerFuncAmplitude(f, f0, Q, Qe, A, theta, slope):
     return np.abs((1.+slope*(f/1.e9-f0)/f0) *
                   HangerFuncAmplitude(f, f0, Q, Qe, A, theta))
 
+def SlopedHangerFuncAmplitudeGuess(data, f, fit_window=None):
+    xvals = f
+    # print(f)
+    # print(np.squeeze(data))
+    peaks = a_tools.peak_finder(xvals, data)
+      # Search for peak
+    if peaks['dip'] is not None:    # look for dips first
+        f0 = peaks['dip']
+        amplitude_factor = -1.
+    elif peaks['peak'] is not None:  # then look for peaks
+        f0 = peaks['peak']
+        amplitude_factor = 1.
+    else:                                 # Otherwise take center of range
+        f0 = np.median(xvals)
+        amplitude_factor = -1.
+        # logging.error('No peaks or dips in range')
+        # If this error is raised, it should continue the analysis but
+        # not use it to update the qubit object
+    min_index = np.argmin(data)
+    max_index = np.argmax(data)
+    min_frequency = xvals[min_index]
+    max_frequency = xvals[max_index]
+    # Fit data according to the model required
+    # added reject outliers to be robust agains CBox data acq bug.
+    # this should have no effect on regular data acquisition and is
+    # only used in the guess.
+    amplitude_guess = max(dm_tools.reject_outliers(data))
+
+    # Creating parameters and estimations
+    S21min = (min(dm_tools.reject_outliers(data)) /
+              max(dm_tools.reject_outliers(data)))
+
+    Q = f0 / abs(min_frequency - max_frequency)
+
+    Qe = abs(Q / abs(1 - S21min))
+    guess_dict = {'f0': {'value': f0*1e-9,
+                         'min': min(xvals)*1e-9,
+                         'max': max(xvals)*1e-9},
+                  'A': {'value': amplitude_guess},
+                  'Q': {'value': Q,
+                        'min': 1,
+                        'max': 50e6},
+                  'Qe': {'value': Qe,
+                         'min': 1,
+                         'max': 50e6},
+                  'Qi': {'expr': 'abs(1./(1./Q-1./Qe*cos(theta)))',
+                         'vary': False},
+                  'Qc': {'expr': 'Qe/cos(theta)',
+                         'vary': False},
+                  'theta': {'value': 0, 
+                            'min': -np.pi/2,
+                            'max': np.pi/2},
+                  'slope': {'value':0,
+                            'vary':True}}
+    return guess_dict
+
 
 def SlopedHangerFuncComplex(f, f0, Q, Qe, A, theta, phi_v, phi_0, slope):
     # This is the function for a hanger (lambda/4 resonator) which takes into
@@ -265,6 +321,10 @@ def linear_with_background_and_offset(x, a, b, c):
     '''
     return np.sqrt((a*x)**2 + b**2)+c
 
+def double_cos_linear_offset(t, amplitude1,amplitude2, frequency1,frequency2, phase1,phase2, slope, offset):
+    cosfn1 = amplitude1*np.cos(2*np.pi*frequency1*t + phase1)
+    cosfn2 = amplitude2*np.cos(2*np.pi*frequency2*t + phase2)
+    return cosfn1 + cosfn2 + offset + slope*t
 
 def gaussian_2D(x, y, amplitude=1,
                 center_x=0, center_y=0,
--- a/pycqed/analysis_v2/base_analysis.py
+++ b/pycqed/analysis_v2/base_analysis.py
@@ -10,12 +10,13 @@ import numpy as np
 
 from matplotlib import pyplot as plt
 from pycqed.analysis import analysis_toolbox as a_tools
+from pycqed.analysis import fitting_models as fit_mods
 import pycqed.analysis_v2.default_figure_settings_analysis as def_fig
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 import datetime
 # import SSRO_generic
 import json
-
+import lmfit
 gco = a_tools.get_color_order
 
 
@@ -191,6 +192,7 @@ class BaseDataAnalysis(object):
         self.extract_data()
         self.process_data()
         if self.do_fitting:
+            self.prepare_fitting()
             self.run_fitting()
         self.prepare_plots()
         if not self.extract_only:
@@ -202,8 +204,38 @@ class BaseDataAnalysis(object):
     def prepare_plots(self):
         pass
 
-    def run_fitting(self):
+    def prepare_fitting(self):
         pass
+
+    def run_fitting(self):
+        # This function does the fitting and saving of the parameters
+        # based on the fit_dict options.
+        # Only model fitting is implemented yet. Minimizing fitting should
+        # be implemented here.
+        fit_guess_fn = self.fit_dict['fit_guess_fn']
+        fit_fn = self.fit_dict['fit_fn']
+        fit_yvals = self.fit_dict['fit_yvals']
+        fit_xvals = self.fit_dict['fit_xvals']
+        model = lmfit.Model(fit_fn)
+        if self.do_timestamp_blocks:
+            self.fit_dict['fit_results'] = []
+            for tt in range(len(fit_yvals)):
+                guess_dict = fit_guess_fn(fit_yvals[tt],
+                                          **fit_xvals[tt])
+                for key, val in list(guess_dict.items()):
+                    model.set_param_hint(key, **val)
+                params = model.make_params()
+                self.fit_dict['fit_results'].append(model.fit(fit_yvals[tt],
+                                                              params=params,
+                                                              **fit_xvals[tt]))
+        else:
+            guess_dict = fit_guess_fn(fit_yvals, **fit_xvals)
+            for key, val in list(guess_dict.items()):
+                    model.set_param_hint(key, **val)
+            params = model.make_params()
+            self.fit_dict['fit_results'] = model.fit(fit_yvals,
+                                                     params=params,
+                                                     **fit_xvals)
 
     def save_figures(self, savedir='', savebase=None, tag_tstamp=True,
                      fmt='pdf', key_list='auto'):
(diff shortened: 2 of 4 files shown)


pycqed/analysis/fitting_models.py
```diff
@@ -237,6 +237,62 @@ def SlopedHangerFuncAmplitude(f, f0, Q, Qe, A, theta, slope):
     return np.abs((1.+slope*(f/1.e9-f0)/f0) *
                   HangerFuncAmplitude(f, f0, Q, Qe, A, theta))
 
+def SlopedHangerFuncAmplitudeGuess(data, f, fit_window=None):
+    xvals = f
+    # print(f)
```
AdriaanRol: Please remove commented out code

pycqed/analysis/fitting_models.py
```diff
@@ -237,6 +237,62 @@ def SlopedHangerFuncAmplitude(f, f0, Q, Qe, A, theta, slope):
     return np.abs((1.+slope*(f/1.e9-f0)/f0) *
                   HangerFuncAmplitude(f, f0, Q, Qe, A, theta))
 
+def SlopedHangerFuncAmplitudeGuess(data, f, fit_window=None):
```
AdriaanRol: Good to have this as a model 👍

pycqed/analysis/fitting_models.py
```diff
@@ -237,6 +237,62 @@ def SlopedHangerFuncAmplitude(f, f0, Q, Qe, A, theta, slope):
...
+    amplitude_guess = max(dm_tools.reject_outliers(data))
+
+    # Creating parameters and estimations
+    S21min = (min(dm_tools.reject_outliers(data)) /
+              max(dm_tools.reject_outliers(data)))
+
+    Q = f0 / abs(min_frequency - max_frequency)
+
+    Qe = abs(Q / abs(1 - S21min))
+    guess_dict = {'f0': {'value': f0*1e-9,
```
AdriaanRol: I think there should not be factors of 1e-9 (or 1e9) at this level. 
As this is mostly a historical thing I suggest we leave it for now but to prevent nasty bugs resulting from conversion we should always use SI units in the code. Special plotting and printing functions exist to ensure that when we display it we get the right prefixes. 

For now I suggest we leave it but you should know about this.

pycqed/analysis_v2/base_analysis.py
```diff
@@ -10,12 +10,13 @@
...
 from matplotlib import pyplot as plt
 from pycqed.analysis import analysis_toolbox as a_tools
+from pycqed.analysis import fitting_models as fit_mods
 import pycqed.analysis_v2.default_figure_settings_analysis as def_fig
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 import datetime
 # import SSRO_generic
 import json
-
+import lmfit
```
AdriaanRol: please removed commented out code

pycqed/analysis_v2/base_analysis.py
```diff
@@ -66,7 +68,6 @@ def get_timestamps(self):
             self.t_stop
             self.labels # currently set from options dict
         """
-        print(self.labels)
```
AdriaanRol: let's remove debug print statemetns.

pycqed/analysis_v2/base_analysis.py
```diff
@@ -188,6 +192,7 @@ def run_analysis(self):
         self.extract_data()
         self.process_data()
         if self.do_fitting:
+            self.prepare_fitting()
```
AdriaanRol: 👍  Good addition

pycqed/analysis_v2/base_analysis.py
```diff
@@ -199,9 +204,39 @@ def process_data(self):
     def prepare_plots(self):
         pass
 
-    def run_fitting(self):
+    def prepare_fitting(self):
```
AdriaanRol: Adding a docstring here would be very helpful. (Similar to what there is on my branch)

pycqed/analysis_v2/base_analysis.py
```diff
@@ -199,9 +204,39 @@ def process_data(self):
...
         pass
 
+    def run_fitting(self):
+        # This function does the fitting and saving of the parameters
```
AdriaanRol: If you start and end comments with triple quotes (''') they will show up as docstrings when you do ? on the method or when (if we ever do this) auto generate docs in the future. 

👍  for the comments

pycqed/analysis_v2/base_analysis.py
```diff
@@ -199,9 +204,39 @@ def process_data(self):
...
+        # based on the fit_dict options.
+        # Only model fitting is implemented yet. Minimizing fitting should
+        # be implemented here.
+        fit_guess_fn = self.fit_dict['fit_guess_fn']
```
AdriaanRol: I thought a bit about this and I think `fit_dicts` is the right name for this. 
However, I would make this a dict of dicts for each fit that one wants to do on the dataset. 

For example, in the flipping analysis I'm making now I want to do both a fit to a cosine (or a decaying cosine) and to a polynomial. 

I would like the fit dict to have the following structure. 
```python 
self.fit_dict{'cos_fit:  {
                                    'fit_guess_fn': some_fun, 'fit_xvals': fit_xval}
                    'poly_fit': {'fit_guess_fn': some_other_fun , .... } 
                    } 
```

pycqed/analysis_v2/base_analysis.py
```diff
@@ -199,9 +204,39 @@ def process_data(self):
...
+        fit_fn = self.fit_dict['fit_fn']
+        fit_yvals = self.fit_dict['fit_yvals']
+        fit_xvals = self.fit_dict['fit_xvals']
+        model = lmfit.Model(fit_fn)
+        if self.do_timestamp_blocks:
```
AdriaanRol: I still don't understand how do timestamp blocks works but I guess that will come.

pycqed/analysis_v2/base_analysis.py
```diff
@@ -199,9 +204,39 @@ def process_data(self):
...
+            self.fit_dict['fit_results'] = []
+            for tt in range(len(fit_yvals)):
+                guess_dict = fit_guess_fn(fit_yvals[tt],
+                                          **fit_xvals[tt])
+                for key, val in list(guess_dict.items()):
+                    model.set_param_hint(key, **val)
+                params = model.make_params()
+                self.fit_dict['fit_results'].append(model.fit(fit_yvals[tt],
```
AdriaanRol: I'm confused that the guess happens here (in the prepare fitting method) I would expect this to happen in the actual fitting method. Especially as it is usally the piece of code that takes most space. 

Otherwise the dictionary doesn't really add much in respect of enforcing structure.

pycqed/analysis_v2/base_analysis.py
```diff
@@ -531,6 +566,7 @@ def plot_color2D(self, pfunc, pdict, axs):
         plot_zrange = pdict.get('zrange', None)
         plot_yrange = pdict.get('yrange', None)
         plot_xrange = pdict.get('xrange', None)
+        plot_xwidth = pdict.get('xwidth',None)
```
AdriaanRol: Not really clear what this parameter does, could you add a comment?

pycqed/analysis_v2/base_analysis.py
```diff
@@ -540,25 +576,37 @@ def plot_color2D(self, pfunc, pdict, axs):
...
 
         if plot_zrange is not None:
             fig_clim = plot_zrange
         else:
             fig_clim = [None, None]
 
+
         if self.do_timestamp_blocks:
-            for tt in len(plot_zvals):
+            for tt in range(len(plot_zvals)):
```
AdriaanRol: I guess this fixes a silly bug.

pycqed/scripts/NanowireAnalysis.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
```
AdriaanRol: The pycqed/scripts folder no longer exists in the master. 
Furthermore, the name NanoWireAnalysis is a project name rather than one related to functionality. 

I would suggest moving this file  to the pycqed_scripts repository (all import statements should work from there so it should not really change anything). If at some point this turns out to be of general use it should be added to the main repository. (I think the "spectroscopy" file below is a good example of that.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
+"""
```
AdriaanRol: I would suggest moving this file to `pycqed/analysis_v2/spectroscopy_analysis.py`.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+import pycqed.analysis_v2.base_analysis as ba
+import numpy as np
+import pandas as pd
+from pycqed.analysis import measurement_analysis as MA
+from pycqed.analysis import analysis_toolbox as a_tools
+from pycqed.analysis.tools import data_manipulation as dm_tools
+from pycqed.analysis import fitting_models as fit_mods
+import lmfit
+
+class Spectroscopy(ba.BaseDataAnalysis):
```
AdriaanRol: A docstring describing that (if) this is an abstract base class and what it is used for/how it should be used would be quite helpful here.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+
+
+
+
+
+    def prepare_plots(self):
+
+        if len(self.data_dict['timestamps']) == 1:
```
AdriaanRol: I guess this is how you deal with "single_timestamp" analyses. I'll see if I can adopt that in some better way. I ran into quite some trouble with the default plot_line functionality.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+            plot_fn = self.plot_line
+            self.plot_dicts['amp'] = {'plotfn': plot_fn,
+                            'xvals': self.plot_frequency,
+                            'yvals': self.plot_amp,
+                            'title': 'Spectroscopy amplitude: %s'%(self.timestamps[0]),
```
AdriaanRol: I suggest we avoid any references to `self."some attribute"` if it can be avoided. In this case the timestamps also exist in the data_dict. 

As a matter of fact (I see more of these above) I would avoid using attribute assignment in the self. completely and stuck exclusively to the data_dict.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+                            'xvals': self.plot_frequency,
+                            'yvals': self.plot_phase,
+                            'title': 'Spectroscopy phase: %s'%(self.timestamps[0]),
+                            'xlabel': self.plot_xlabel,
+                            'ylabel': self.phase_label,
+                            'yrange': self.phase_range,
+                            'plotsize': self.plotsize
+                            }
+        else:
+            print('plotting not yet coded up for multiple traces')
```
AdriaanRol: replace `print` with `logging.warning` (may need `import logging` at the top of the file)

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+        self.amp_label = 'Transmission amplitude (V rms)'
+        self.phase_label = 'Transmission phase (degrees)'
+        if len(self.data_dict['timestamps']) == 1:
+            self.plot_phase = np.unwrap(np.pi/180.*self.plot_phase)*180/np.pi
+            self.plot_xlabel = 'Readout Frequency (GHz)'
+        else:
+            pass
+
+    def prepare_fitting(self):
+        # Fitting function for one data trace. The fitted data can be
```
AdriaanRol: use triple quotes to specify docstrings.

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+                             'fit_yvals': self.plot_amp,
+                             'fit_xvals': {'f': self.plot_frequency}
+                             }
+        else:
+            self.fit_dict = {'fit_fn': fit_fn,
+                             'fit_guess_fn': fit_guess_fn,
+                             'fit_yvals': [np.squeeze(tt) for tt in self.plot_amp],
+                             'fit_xvals': np.squeeze([{'f': tt[0]} for tt in self.plot_frequency])}
+
+    def do_subtract_background(self,thres=None,back_dict=None,):
```
AdriaanRol: Maybe a short docstring on how it does this

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+        if len(self.data_dict['timestamps']) == 1:
+            pass
+        else:
+            x_filtered = []
+            y_filtered = []
+            for tt in range(len(self.data_dict['timestamps'])):
+                y = np.squeeze(self.plot_amp[tt])
+                x = np.squeeze(self.plot_frequency)[tt]
+                # print(self.plot_frequency)
```
AdriaanRol: please remove commented out code

pycqed/scripts/Spectroscopy.py
```diff
@@ -0,0 +1,281 @@
...
+                                 **fit_xvals)
+            self.background_fit = fit_res
+
+            for tt in range(len(self.data_dict['timestamps'])):
+                divide_vals = fit_fn(np.squeeze(self.plot_frequency)[tt]*1e-9,**fit_res.best_values)
+                self.plot_amp[tt] = np.array([np.array([np.divide(np.squeeze(self.plot_amp[tt]),divide_vals)])]).transpose()
+
+
+
+    def plot_fitting(self):
```
AdriaanRol: I specified these plots in the same way as normal plots using the plotting dicts. 
I will try to make my example a bit more clear but I think that method is better as it enforces the same paradigm everywhere.

PR #301, opened 2017-08-03
Removed exception catching that fails in python <3.6

elrama-:
Motivated this pull request:
- Versions of python before 3.6 fail to catch the exception ModuleNotFoundError.
- Tested on two computers running PycQED but no ZI API (that was the module that failed).

- Removed an exception catching

Comments:
- I left the original line commented. We might be able to implement this once all the computers upgrade to python 3.6 (I don't think that is yet reasonable to ask).

Reviewers? @AdriaanRol


AdriaanRol:
@elrama- Could you give a status update on this?

PR #308, opened 2017-08-28
Pulse Scheme Plotting

lciorciaro:
This PR contains some code I wrote to plot pulse schemes.
Features:
- Create a figure with axes and frame disabled to draw the pulse scheme on
- MW pulse (modulated Gaussian) with optional label
- Simple flux pulse (smoothed square pulse with shaded area)
- Labeling an interval with arrows, optional vertical lines and an optional text label

A simple example for the usage is included.
It should be quite straight forward to add new pulse shapes as needed.

### DIFF
--- a/examples/pulse_scheme_example.py
+++ b/examples/pulse_scheme_example.py
@@ -1,0 +1,24 @@
+import matplotlib.pyplot as plt
+import pycqed.utilities.pulse_scheme as pls
+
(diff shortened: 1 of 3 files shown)


lciorciaro:
@AdriaanRol added example as test.

Here is an example output:

pulse_scheme_example

PR #309, opened 2017-09-07
Enh/ccl driver

AdriaanRol:
- Moved over the CCL driver from the ElecProj_CCL repository 
- Tested and ran the example 
- There are possible improvements to the driver (listed in issue # xx) 

@kel85uk @gtaifu


AdriaanRol:
@nyxus , I have gone through the CCL driver in quite some detail. It appears the auto generating of parameter files that we have discussed in the context of the DDM and QWG is done quite nicely here. I think it would pay of to take a closer look at this. Let me know if I need to show you in some more details.

gtaifu:
@AdriaanRol 
It seems that some code refining is required before merging. Can you do that?
image

image

AdriaanRol:
OK, will do

PR #319, opened 2017-09-18
Merging the starmon branch

AdriaanRol:
Because of the massive amount of changes on the starmon branch, most of which are relevant to current experiments, it has become both quite important as well as quite hard to merge these into the master using small self contained pull requests. 

After discussing with @brianzi , we think the best course of action is to have a single pull request containing all the changes on the starmon branch. The review process will consist of @brianzi using the code in the starmon branch. If he finds that this works as intended we conclude that this is safe to merge. If anyone objects to this please propose a better alternative. I will also try to give an overview of the changes in the Starmon branch below. 

- Cryo scope sequences and relevant analysis
- QWG Lookuptable managers 
- GST for CBox
- CBox sequences (and a lot of them) 
- Device object (basic first attempt) 

And of course loads of other changes. 

Again, none of this should break anything but let's make sure by testing.


AdriaanRol:
@brianzi , You have been using this for a while now. As I understand there are no serious problems. Could you give a status update?

PR #325, opened 2017-09-21
bug:AWG start order

antsr:
Fixes a problem when using multiple AWGs where the slave AWGs might be started later than the master AWG resulting in a situation, where readout instrument is triggered (from master AWG), but the drive pulses are not output (as the slave AWG is not yet running). With the old implementation, the problem appears quite rarely and is therefore difficult to detect.

- Instead of waiting a predetermined time before starting the master AWG, ask the slave AWGs if they are running and start the master only if they are.

Could be reviewed by someone using multiple AWGs at DCL.


pycqed/measurement/waveform_control/pulsar.py
```diff
@@ -189,8 +189,20 @@ def start(self):
             for AWG in self.used_AWGs():
                 if AWG != self.master_AWG():
                     self._start_AWG(AWG)
-            time.sleep(0.2)  # wait 0.2 second for all other awg-s to start
-            self._start_AWG(self.master_AWG())
+            tstart = time.time()
+            for AWG in self.used_AWGs():
+                if AWG != self.master_AWG():
```
AdriaanRol: A comment here indicating why the master should only be started once the rest is already running would be good. 

A similar comment should be added in line 204.

pycqed/measurement/waveform_control/pulsar.py
```diff
@@ -532,6 +544,15 @@ def _stop_AWG(self, AWG):
         else:
             raise ValueError('Unsupported AWG type: {}'.format(type(obj)))
 
+    def _is_AWG_running(self, AWG):
+        obj = self.AWG_obj(AWG=AWG)
+        if isinstance(obj, Tektronix_AWG5014):
+            return obj.get_state() != 'Idle'
+        elif isinstance(obj, UHFQC):
```
AdriaanRol: If I understand correctly this will directly break the use of the Pulsar for use with any UHFQC. As far as I know you guys are the only ones using that but it might be worth checking that.

pycqed/measurement/waveform_control/pulsar.py
```diff
@@ -532,6 +544,15 @@ def _stop_AWG(self, AWG):
...
+            raise NotImplementedError()
+        else:
+            raise ValueError('Unsupported AWG type: {}'.format(type(obj)))
```
AdriaanRol: I'd say that Unsupported AWG type should be a NotImplementedError (with exactly that message). It may also be beneficial to add the name of the object so that people know which of their AWG's is of the offending type. 

Also worth checking if classes that inherit from this (say maybe some other AWG5014 sub type) would raise an error here. I'm mostly thinking of this in the context of dummy instruments.

antsr: The `else` case is meant for the case when some object is (accidentally) specified as the AWG for the Pulsar that is not meant to be supported. So I consider `ValueError` more appropriate than `NotImplementedError`.

pycqed/measurement/waveform_control/pulsar.py
antsr: Yes, it is true that this will break support for the UHFQC as a slave AWG. But since we are not using it currently, I prefer to leave it unimplemented rather than implement something I could not immediately test.

AdriaanRol:
@antsr I approve of your changes. I will OK for merging once it has been tested on a setup with an AWG5014 here.

PR #330, opened 2017-09-27
Improvements to measurement_analysis.py and analysis_toolbox.py.

### LINKED ISSUE #327: Homodyne analysis creates ugly plots
frequency_frequency hz _vs_mag
Compare labels and size :
frequency_x_vs_y

Vintrix:
Improves inheritance tree, plotting, and some analysis routines in measurement_analysis.py. Improves peak_finder and extends rotate_and_normalize_data based on calibration points in analysis_toolbox.by. Details about the changes made to each module are given below. 
 

Motivation: 
* Several pieces of code, like plotting or call to cal_points calibration method, were duplicated across different analysis routines instead of inheriting from base classes. This produced inconsistent results and plots 
* For the analysis classes listed below, some of the plots were not well-proportioned or would not fit inside the saved PNG figure
* peak_finder would often fail to find good guesses, or analysis routines would fail because peak_finder returned Nones
 
Fixes #327.

# measurement_analysis.py
Improves
* The following classes
   * MeasurementAnalysis (MA) base
      * Improves main plotting routine _plot_results_vs_sweepparam and includes it in all classes above so that plotting is consistent.
      * Now has attributes for plot line widths and font sizes defined in the _init_, which are used in _plot_results_vs_sweepparam_
      * Adds to default_fig PRL standard figure sizes for 1 or 2 columns, which can be chosen by setting "no_of_columns"
      * Adds method scale_sweep_points that creates "self.scaled_sweep_points", "self.scaled_xlabel", "self.scaled_sweep_unit." These contain correctly scaled values for each experiment (ex: GHz instead of Hz for spectroscopy, us instead of s for Ramsey or T1)
   * Qubit_Spectroscopy_Analysis
      * Adds support for finding the second peak at f_gf/2 from a high power spectroscopy measurement
      * Improves fitting by improving the _analysis_toolbox/peak_finder_ routine to give more accurate parameter guesses 
      * **! Uses the old way of calculating data_dist based on data_amp and data_phase, but the latest version from the origin/master is there and commented out.**
   * Homodyne_Analysis
      * Improves plotting
      * Uses improved  _analysis_toolbox/peak_finder_
   * SSRO_Analysis
      * reorganize functions inside the analysis: make 2D histogram plotting an independent subroutine
      * make plotting optional as it is the slowest part for small datasets
      * improve the look of the 1D histogram plot
      * add support for leaving out part of the data in the final analysis depending on the result of preselection readout (an additional measurement before the state preparation pulse)
 
   * TD_Analysis (TD) base for all time-domain classes
      * New input parameters: 
         * _for_ef_: perform analysis for the EF transition/ on the second excited state
         * _last_ge_pulse_: for experiments on the EF transition, whether or not a GE pi-pulse was applied before each readout, placing qubit in ground state
         * _start_at_zero_: This parameter is used in _TD.normalize_data_to_calibration_points_ to ensure that the data trace after calibration to cal_points starts at (closest to) zero. Currently only used for _Rabi_Analysis_ to force the cosine data trace to have zero phase (allows to set phase to 0 during fit). 
      * Adds support for 0, 2, 4, or 6 cal_points in _TD.normalize_data_to_calibration_points_
         * 6 cal_points are used in EF measurements. Based on the "last_ge_pulse" parameter, normalize_data_to_calibration_points picks the relevant 4 cal_points from those 6 (I and X180_ef calibration points if last_ge_pulse else X180 and X180_ef calibration points) and sends them to analysis_toolbox/rotate_and_normalize_data
         * 0 or 2 (just I pulses) cal points are used for Rabi on GE because the X180 pulse is typically not calibrated before a Rabi measurement
   * Rabi_Analysis
      * Implements complete Rabi analysis: calibration based on cal_points, result is fitted to cosine model, fit results are plotted, pi-pulse and pi/2-pulse amplitudes are extracted
   * Ramsey_Analysis
      * Adds support for two values of the artificial_detuning keyword argument which must be passed in
      * Now calculates the new qubit frequency based on the old value and the fit results
   * Qscale_Analysis
      * Implements analysis for a DRAG pulse calibration measurement as described in Baur, M. PhD Thesis(2012), and extracts optimal qscale value
   * T1_Analysis
      * No major improvements other than plotting and inheritance structure, as described under "Structure of measurement_analysis module" below
 
**! All time-domain routines above return parameters of interest, self.parameter_name, as DICTIONARIES with keys "parameter_name" and "parameter_name_stderr."** Example: self.T1 = {'T1':T1, 'T1_stderr':T1_stderr}. This was done to allow to save them in the HDF file, using the new _MA.save_computed_parameters_.
 
* Structure of measurement_analysis module 
   * Improves inheritance structure for the classes above: both _init_ and _run_default_analysis_ of specific analysis classes follow the trees: (specific time-domain classes -> TD -> MA) or 
(Qubit_Spectroscopy_Analysis/Homodyne_Analysis -> MA). Hence, plot of measured_values vs sweep_points is always done in _MA.run_default_analysis()_, and time-domain classes (except Qscale and Ramsey for 2 artificial detunings) plot fit results on self.fig and self.ax created in TD, which just plots data after calibration based on cal_points.
   * Each of the classes above takes "qb_name" which allows to retrieve previously measured values of the parameters of interest and display them in the text boxes underneath the plots. **Ramsey_Analysis must be given qb_name to retrieve last measured value of qubit frequency in order to correctly calculate new qubit frequency.**
 
# analysis_toolbox.py
Improves
* peak_finder
   * If the parameter optimize==False (default), identical to old peak_finder routine (now renamed to look_for_peaks_dips), which was also improved
   * If optimize==True, reruns look_for_peaks_dips until it finds tallest peak/smallest dip in the data set
   * This does not break usage of peak_finder as the old routine throughout measurement_analysis.py
* rotate_and_normalize_data
   * Adds support for 0 and 2 calibration points

Adds
* look_for_peaks_dips
   * Original peak_finder improved and renamed
* find_second_peak
   * Finds second largest peak/smallest dip in a data set; used for finding f_gf/2 in Qubit_Spectroscopy_Analysis
* rotate_and_normalize_data_no_cal_points
 
# fitting_models.py
Changes
* TwinLorentzFunc input parameter names renamed

@AdriaanRol

### DIFF
--- a/pycqed/analysis/analysis_toolbox.py
+++ b/pycqed/analysis/analysis_toolbox.py
@@ -1,11 +1,13 @@
 # some convenience tools
 #
+import inspect
 import logging
 import numpy as np
 import os
 import time
 import datetime
 import warnings
+from copy import deepcopy
 from collections import OrderedDict as od
 from matplotlib import pyplot as plt
 from matplotlib import colors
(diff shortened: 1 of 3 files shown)


AdriaanRol:
@Vintrix thanks for your contribution, the measurement analysis is indeed something that is in need of some improvements. However, I have some general comments to this PR. 

- There are conflicts with the master. Please resolve these. I will not look into the code before this is done. 
- Codacy is highlighting issues, please resolve these. 

- The PR is large, making it very hard to review. If possible please split this PR up into multiple smaller ones. 
- There are no tests. I am paranoid about accepting changes that can break other things.  (such as those involving parameter name changes). The measurement analysis code is quite easy to test as it does not require a setup. Please add tests to your code. 

A lot of the changes you propose are already part of measurement analysis v2 (which can be found on the master). Especially ideas such as keeping several options in dictionaries. The analysis v2 has the ambition to structurally solve these problems and be more documented. 

Now let me go over your points (note I'm not looking at the code yet as there are still many conflicts) 

 

>Motivation: 
>* Several pieces of code, like plotting or call to cal_points calibration method, were duplicated across different analysis routines instead of inheriting from base classes. This produced inconsistent results and plots 

This is correct and should indeed be fixed, it is not as simple as simply deleting the old versions as there is legacy code that uses that. A single pull request containing only those changes I would most likely approve. 

> * For the analysis classes listed below, some of the plots were not well-proportioned or would not fit inside the saved PNG figure

I have no such experience but again I think it is good to address this. If you did, please provide comparison figures (you can just drag them into the PR on github) to show what changed. 
> * peak_finder would often fail to find good guesses, or analysis routines would fail because peak_finder returned Nones
I have experienced similar issues, a single pull request containing only that change I would most likely approve. Adding a test would also be appreciated
 
> Fixes #327.

I'm not sure it does, are you using the unit aware axis relabeling function? I get the impression you do most of this by hand. 

> Changes proposed in this pull request:

> # measurement_analysis.py
Improves
> * The following classes
>   * MeasurementAnalysis (MA) base
>      * Improves main plotting routine _plot_results_vs_sweepparam and includes it in all classes above so that plotting is consistent.

:+1: 
>      * Now has attributes for plot line widths and font sizes defined in the _init_, which are used in _plot_results_vs_sweepparam_
>      * Adds to default_fig PRL standard figure sizes for 1 or 2 columns, which can be chosen by setting "no_of_columns"
>      * Adds method scale_sweep_points that creates "self.scaled_sweep_points", "self.scaled_xlabel", "self.scaled_sweep_unit." These contain correctly scaled values for each experiment (ex: GHz instead of Hz for spectroscopy, us instead of s for Ramsey or T1)

Are you aware of the unit aware set_xlabel and set_ylabel? They do this automatically. A push to use SI units everywhere is related to this. 
>   * Qubit_Spectroscopy_Analysis
>      * Adds support for finding the second peak at f_gf/2 from a high power spectroscopy measurement

I am not sure if this should be a separate class or added to the base spectroscopy analysis. I know that @tstavenga has a lot of changes for the spectroscopy analysis in measurement analysis v2 that are not yet integrated into the master. 
>      * Improves fitting by improving the _analysis_toolbox/peak_finder_ routine to give more accurate parameter guesses 

As a separate PR I would most likely approve this. Are you aware of the use of the `guess` method and the way these model guesses can be made to be independent of the specific analysis? 


>      * **! Uses the old way of calculating data_dist based on data_amp and data_phase, but the latest version from the origin/master is there and commented out.**

Sounds like a bug that you fixed
>   * Homodyne_Analysis
>      * Improves plotting
>      * Uses improved  _analysis_toolbox/peak_finder_

Sounds good but hard to evaluate due to the size of this PR

>   * SSRO_Analysis
>      * reorganize functions inside the analysis: make 2D histogram plotting an independent subroutine
>      * make plotting optional as it is the slowest part for small datasets
>      * improve the look of the 1D histogram plot
>      * add support for leaving out part of the data in the final analysis depending on the result of preselection readout (an additional measurement before the state preparation pulse)
 
This analysis is in dire need of a redo so these are probably very good changes, again we should split it up. 

>   * TD_Analysis (TD) base for all time-domain classes
>      * New input parameters: 
>         * _for_ef_: perform analysis for the EF transition/ on the second excited state

Even though not clear from the name, the TD analysis is intended mostly for single qubit time domain experiments (the straightforward ones). As such I think including this option adds unneeded complexity. Furthermore it is not clear to me from your description what this would do. 

>         * _last_ge_pulse_: for experiments on the EF transition, whether or not a GE pi-pulse was applied before each readout, placing qubit in ground state
>         * _start_at_zero_: This parameter is used in _TD.normalize_data_to_calibration_points_ to ensure that the data trace after calibration to cal_points starts at (closest to) zero. Currently only used for _Rabi_Analysis_ to force the cosine data trace to have zero phase (allows to set phase to 0 during fit). 

This sounds like magic applied in the analysis that is unneeded and will confuse people. 

>      * Adds support for 0, 2, 4, or 6 cal_points in _TD.normalize_data_to_calibration_points_

This is already there or I am misunderstanding what you mean here. 


I haven't had time yet to read the rest of your comments. I'll try to do that this week. 

Best,

Vintrix:
@AdriaanRol thanks for all the comments! And thanks for drawing my attention to the unit aware set_xlabel and set_ylabel. I have updated my code to use these.

I am now working on splitting up this PR into smaller PRs, where I will also include tests.

One quick clarification the TD_Analysis. I have also thought of TD_Analysis as a base for single qubit time domain experiments, but I also included qutrit experiments (for example performing Rabi between |e> and |f> states) since the analysis routines are basically identical. "for_ef" is there to put correct yaxis labels, and to retrieve previously measured values for the correct parameters from the HDF file (amp180_ef instead of amp180, for example).

Best,

AdriaanRol:
Closing this PR as it has been split up in several new PR's.

PR #333, opened 2017-09-29
Qwg add trigger logic input parameter

nyxus:
Issue number: https://github.com/DiCarloLab-Delft/QuTech_AWGmodule/issues/47

- Add a new parameter to the QWG to readout the current trigger input logic 

@AdriaanRol


pycqed/instrument_drivers/physical_instruments/QuTech_AWG_Module.py
```diff
@@ -182,6 +182,22 @@ def add_parameters(self):
...
 
+        # Trigger parameters
+        doc_trgs_log_inp = 'Reads the current input values on the all the trigger ' \
+                    +'inputs.\nReturn:\n    uint32 where trigger 1 (T1) ' \
+                    +'is on the Least significant bit (LSB), T2 on the second  ' \
+                    +'bit after LSB, etc.\n\n For example, if only T3 is ' \
+                    +'connected to a high signal, the return value is: ' \
+                    +'4 (0b0000100)\n\n Note: To convert the return value ' \
+                    +'to a readable ' \
+                    +'binary output use: `print(\"{0:#010b}\".format(qwg.' \
```
AdriaanRol: Would it be a suggestion to do this conversion in the `get_parser`? That way the return value is always readable.

nyxus: I thought of doing the conversion via the `get_parser`, but if we like to use the triggers in some kind of logic it is easier to (in my opinion) to have an uint instead of string. With the string, you probably need to do additional parsing (back to an int).

AdriaanRol: If you add a comment explaining that reasoning I'll approve the PR :+1:

Issue #226, opened 2017-05-19
Bug with axis labels function when automatic rescaling happens


AdriaanRol:
The neat axis label functions introduced in #201 (issue #176) shows wrong labels when the axis is rescaled. To be more explicit: I think the problem is that it is not able to adapt if the number or location of ticks is changed.

AdriaanRol:
A workaround to this problem is to always recall the set_xaxis and set_yaxis methods after any rescaling of the figure/before saving.

PR #341, opened 2017-10-02
Enhancement/analysis v1 plotting

### LINKED ISSUE #327: Homodyne analysis creates ugly plots

Vintrix:
Improve plotting in measurement_analysis.py and analysis_toolbox.py. Improves MeasurementAnalysis.plot_results_vs_sweepparams() and color_plot in analysis_toolbox. Improves plotting in Homodyne_Analysis.
 
Motivation: Some of the plots were not well-proportioned or would not fit inside the saved PNG figure

Fixes #327 .

* measurement_analysis.py
   * Add to MeasurementAnalysis.default_fig() the PRL standard figure sizes for 1 or 2 columns, which can be chosen by setting "no_of_columns"
   * Add attributes "dpi", "line_width", "font_size", "tick_length", "tick_width", "marker_size" to MeasurementAnalysis (MA) base 
   * Add the above attributes to MA.plot_results_vs_sweepparam() and to the Homodyne_Analysis class 
* analysis_toolbox.py
   * Add the unit aware set_xlabel and set_ylabel to color_plot used in MA.run_default_anlaysis(TwoD=True)
   * Add plot_errorbars
* test_spectroscopy_analysis.py
   * Add tests for the Hanger model in Homodyne_Analysis

@AdriaanRol 

Master version:
master_version

Improved version:
improved

Master version:
master_version

Improved version:
improved


pycqed/analysis/analysis_toolbox.py
```diff
@@ -1352,13 +1353,13 @@ def color_plot_slices(xvals, yvals, zvals, ax=None,
             zvals[xx] = np.log(zvals[xx])/np.log(10)
 
     # add blocks to plot
-    hold = kw.pop('hold', False)
+    #hold = kw.pop('hold', False)
```
AdriaanRol: Why is this commented out?

pycqed/analysis/analysis_toolbox.py
```diff
@@ -1341,7 +1342,7 @@ def color_plot_slices(xvals, yvals, zvals, ax=None,
     # various plot options
     # define colormap
     cmap = kw.pop('cmap', 'viridis')
-    clim = kw.pop('clim', [None, None])
+    #clim = kw.pop('clim', [None, None])
```
AdriaanRol: Why is this commented out?

pycqed/analysis/analysis_toolbox.py
```diff
@@ -1278,17 +1279,17 @@ def color_plot(x, y, z, fig, ax, cax=None,
     if plot_title is not None:
         ax.set_title(plot_title, y=1.05, fontsize=18)
     if transpose:
-        ax.set_xlabel(ylabel)
-        ax.set_ylabel(xlabel)
+        set_xlabel(ax, ylabel, unit=y_unit)
```
AdriaanRol: Related to #226 , it is better to set the lables after setting the axis limits or all the way at the end of the plotting. This is only a workaround but does do the job we expect it to.

pycqed/analysis/analysis_toolbox.py
```diff
@@ -1352,13 +1353,13 @@ def color_plot_slices(xvals, yvals, zvals, ax=None,
...
     # add blocks to plot
-    hold = kw.pop('hold', False)
+    #hold = kw.pop('hold', False)
     for xx in range(len(xvals)):
         tempzvals = np.array([np.append(zvals[xx], np.array(0)),
                               np.append(zvals[xx], np.array(0))]).transpose()
-        im = ax.pcolor(xvertices[xx:xx+2],
-                       yvertices[xx],
-                       tempzvals, cmap=cmap)
+        # im = ax.pcolor(xvertices[xx:xx+2],
```
AdriaanRol: Again, why is this commented?

pycqed/analysis/analysis_toolbox.py
```diff
@@ -1431,6 +1432,24 @@ def color_plot_interpolated(x, y, z, ax=None,
         return ax, CS, cbar
     return ax, CS
 
+def plot_errorbars(x, y, ax=None, linewidth=2 ,markersize=2, marker='none'):
```
AdriaanRol: Sensible addition.

pycqed/analysis/measurement_analysis.py
```diff
@@ -47,15 +49,30 @@
...
         if folder is None:
             self.folder = a_tools.get_folder(**kw)
         else:
             self.folder = folder
         self.load_hdf5data(**kw)
         self.fit_results = []
-        self.box_props = dict(boxstyle='Square', facecolor='white', alpha=0.8)
         self.cmap_chosen = cmap_chosen
+        self.no_of_columns = no_of_columns
+        self.dpi = 600                  #dpi for plots
```
AdriaanRol: I don't think hardcoding these here makes sense. 
It would be much better to use `kw.get(default_value, 'dict_key')` instead. 

Even better would be some fancy system (like in ma_v2) but let's not worry abou that now. 

Not that I do think that keeping it in the same place for all analyses (like done here) is a very good idea and definitely an improvement 👍

pycqed/analysis/measurement_analysis.py
```diff
@@ -92,17 +109,27 @@ def analysis_h5data(self, name='analysis'):
...
         if fig is None:
             fig = self.default_fig(*arg, **kw)
         ax = fig.add_subplot(111)
-        ax.set_title(self.timestamp_string+'\n'+self.measurementstring)
+        #ax.set_title(self.timestamp_string+'\n'+self.measurementstring)
         ax.ticklabel_format(useOffset=False)
         return fig, ax
 
-    def save_fig(self, fig, figname=None, xlabel='x', ylabel='y',
+    def save_fig(self, fig, figname=None, xlabel='x', ylabel='measured_vlaues',
```
AdriaanRol: Typo: `measured_vlaues` -> `measured_values`

pycqed/analysis/measurement_analysis.py
```diff
@@ -111,8 +138,11 @@ def save_fig(self, fig, figname=None, xlabel='x', ylabel='y',
             plot_formats = [plot_formats]
         for plot_format in plot_formats:
             if figname is None:
-                figname = (self.sweep_name+'_'+xlabel +
-                           '_vs_'+ylabel+'.'+plot_format)
+                if xlabel=='x':
+                    xlabel=self.sweep_name
+                # figname = (self.sweep_name+'_'+ylabel +
```
AdriaanRol: Please remove commented out code

pycqed/analysis/measurement_analysis.py
```diff
@@ -124,9 +154,10 @@ def save_fig(self, fig, figname=None, xlabel='x', ylabel='y',
                     print('WARNING: Could not set tight layout')
             try:
                 fig.savefig(
-                    self.savename, dpi=300,
+                    self.savename, dpi=self.dpi,format=plot_format,
+                    bbox_inches='tight')
                     # value of 300 is arbitrary but higher than default
-                    format=plot_format)
+                    # format=plot_format, bbox_inches='tight',pad_inches=0.25)
```
AdriaanRol: please remove commented out code.

pycqed/analysis/measurement_analysis.py
```diff
@@ -124,9 +154,10 @@ def save_fig(self, fig, figname=None, xlabel='x', ylabel='y',
                     print('WARNING: Could not set tight layout')
             try:
                 fig.savefig(
```
AdriaanRol: Perhaps we can re-set the set_xlabel and set_ylabel in the savefig to work around #226, as this is also where the rescaling happens (from 'tight')

pycqed/analysis/measurement_analysis.py
```diff
@@ -455,24 +557,86 @@ def get_naming_and_values(self):
...
-    def plot_results_vs_sweepparam(self, x, y, fig, ax, show=False,
-                                   marker='-o',
-                                   xlabel=None, x_unit=None,
-                                   ylabel=None, y_unit=None,
-                                   log=False, label=None, **kw):
-        save = kw.pop('save', False)
-        self.plot_title = kw.pop('plot_title',
-                                 textwrap.fill(self.timestamp_string + '_' +
-                                               self.measurementstring, 40))
+    def scale_sweep_points(self, unit_prefix = '', TwoD=False):
```
AdriaanRol: Why is there a scale sweep points method here? 
I think there should be one and only one such method (the one used in set_xlabel/set_ylabel) that you are also importing in this module. 

Let's avoid multiple functions that do the same thing and the related problems.

pycqed/analysis/measurement_analysis.py
```diff
@@ -3920,36 +4095,36 @@ def run_default_analysis(self, print_fit_results=False,
             Qe = abs(Q / abs(1 - S21min))
 
             # Note: input to the fit function is in GHz for convenience
-            HangerModel.set_param_hint('f0', value=f0*1e-9,
+            Model.set_param_hint('f0', value=f0*1e-9,
```
AdriaanRol: Model value *1e-9 still scares me but I assume this is not solved in this PR.

pycqed/tests/test_spectroscopy_analysis.py
```diff
@@ -0,0 +1,32 @@
...
+    @classmethod
+    def setUpClass(self):
+        self.datadir = os.path.join(pq.__path__[0], 'tests', 'test_data')
+        ma.a_tools.datadir = self.datadir
+
+    def test_Homodyne_analysis_hanger(self):
+        #lorentzian model tested in test_heterodyne_analysis.py
+
+        a = ma.Homodyne_Analysis(timestamp='20170929_174145', label='resonator_spec')
+        try:
```
AdriaanRol: Try except statements in a test don't sound like a very good idea. 
Please choose only one case that passes and do not use try/except. .

pycqed/tests/test_spectroscopy_analysis.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+import pycqed as pq
+import os
+from pycqed.analysis import measurement_analysis as ma
+
+
+class Test_Homodyne_analysis(unittest.TestCase):
```
AdriaanRol: happy you added some tests for this 👍 . 

Are you familiar with `pycqed/tests/test_heterodyne_analysis` (bad name I know)? It does basically the same. 

I think more tests is better in this case (especially since you tests specific models). Could you merge these two files?

pycqed/analysis/analysis_toolbox.py
Vintrix: The lines that I commented out were reported by codacy as "possible new issues" because they were unused variables. They were in functions I have not looked at so I didn't want to completely remove these lines.

pycqed/analysis/measurement_analysis.py
Vintrix: Yes, things like this scare me too, but the reason this is so here and in all my PRs is because the HangerFuncAmplitude and the SlopedHangerFuncAmplitude models were both defined with inconsistent units. I should have changed this but for some reason I didn't get to.

hangerfuncamplitude

slopedhangerfuncamplitude

pycqed/analysis/measurement_analysis.py
Vintrix: I will not add this feature to this PR. The save_fig method is used by all the analysis classes to save all sorts of different types of figures (with or without subplots, TwoD and not TwoD plots, many different axis labels and units). Hence, if I understand this correctly, one has to think about how to make the X/Y labels resetting flexible enough (set_xlabel/set_ylabel need to know the axis, the label, and the unit), and how to add these changes to save_fig such that they do not break existing code. I'm afraid that changing this method here will break routines in PR #347.

pycqed/analysis/measurement_analysis.py
Vintrix: I fixed this in PR #341. Here I will leave this unchanged so there is no overlap with the other PR.

Vintrix:
@AdriaanRol , thanks for all the detailed comments!

PR #347, opened 2017-10-02
Enhancement/analysis v1 time domain

Vintrix:
Improve inheritance tree and some single qubit time domain analysis routines in measurement_analysis.py. Extends rotate_and_normalize_data based on calibration points in analysis_toolbox.by to work for 0 or 2 cal_points as well (the 0 cal_points only works if len(measured_values) == 2).
 
Motivation: 
* Several pieces of code, like plotting or call to cal_points calibration method, were duplicated across different analysis routines instead of inheriting from base classes. This produced inconsistent results and plots.
* Only support for 4 cal_point existed but it is useful to have 0 or 2 (two I points) cal_points for a Rabi measurement which calibrates the X180 pulse amplitude.

* measurement_analysis.py
   * MeasurementAnalysis (MA) base
      * Add attributes "dpi", "line_width", "font_size", "tick_length", "tick_width", "marker_size" to MeasurementAnalysis (MA) base. Here, they are not incorporated into any plotting functions in the base class but the new time domain analysis classes use them.
      * Add "save_computed_parameter()" to save parameters of interest that are calculated from fit results (ex: amp180)
   * TD_Analysis (TD) base for all single qubit and qutrit time-domain classes
      * New input parameters:
         * for_ef: perform analysis for the EF transition/on the second excited state. Only used to 
to put correct yaxis labels, and to retrieve previously measured values for the correct parameters from the HDF file (amp180_ef instead of amp180, for example).
         * last_ge_pulse (kw parameter): for experiments on the EF transition, whether or not a GE pi-pulse was applied before each readout, placing qubit in ground state (see next bullet point).
      * Adds support for 0, 2, 4, or 6 cal_points in TD.normalize_data_to_calibration_points
         * 6 cal_points (two I, 2 X180, and 2 X180_ef pulses) may be used in EF measurements. It is sometimes nice to see where the ground, excited, and second excited states are relative to the data points. Based on the "last_ge_pulse" parameter, normalize_data_to_calibration_points picks the relevant 4 cal_points from those 6 (I and X180_ef calibration points if last_ge_pulse else X180 and X180_ef calibration points) and sends them to analysis_toolbox/rotate_and_normalize_data
         * 0 or 2 (just I pulses) cal points are used for Rabi on GE because the X180 pulse is typically not calibrated before a Rabi measurement. The 0 cal_points only works if len(measured_values) == 2.
   * Rabi_Analysis
      * Implements complete Rabi analysis: calibration based on cal_points, result is fitted to cosine model, fit results are plotted, pi-pulse and pi/2-pulse amplitudes are extracted from fit results.
      * ! This new class is called Rabi_Analysis; the master version was renamed to Rabi_Analysis_original.
   * Ramsey_Analysis
      * Adds support for two values of the artificial_detuning keyword argument which must be passed in
      * Now calculates the new qubit frequency based on the old value and the fit results
   * Qscale_Analysis
      * Implements analysis for a DRAG pulse calibration measurement as described in Baur, M. PhD Thesis (2012), and extracts optimal qscale value.
   * T1_Analysis
      * No major improvements other than plotting and inheritance structure, as described under "Structure of measurement_analysis module" below
 
   ! All time-domain routines above return parameters of interest, self.parameter_name, as DICTIONARIES
   with keys "parameter_name" and "parameter_name_stderr." Example: self.T1 = {'T1':T1, 
   'T1_stderr':T1_stderr}. This was done to allow to save them in the HDF file, using the new 
   MA.save_computed_parameters().
 
   * Structure of measurement_analysis module 
      * Improves inheritance structure for the classes above: both __init__ and run_default_analysis of specific analysis classes follow the trees: (specific time-domain classes -> TD -> MA) or 
(Qubit_Spectroscopy_Analysis/Homodyne_Analysis -> MA). Hence, plot of measured_values vs sweep_points is always done in MA.run_default_analysis(), and time-domain classes (except Qscale and Ramsey for 2 artificial detunings) plot fit results on self.fig and self.ax created in TD, which just plots data after calibration based on cal_points.
      * Each of the classes above takes "qb_name" which allows to retrieve previously measured values of the parameters of interest and display them in the text boxes underneath the plots. Ramsey_Analysis must be given qb_name to retrieve last measured value of qubit frequency in order to correctly calculate new qubit frequency.
 
* analysis_toolbox.py:
   * Add support for 2 calibration points in rotate_and_normalize_data
   * Add rotate_and_normalize_data_no_cal_points for 0 cal_points
 
* test_timedomain_analysis.py
   * Add test for new Rabi_Analysis class; rename original test for rabi analysis to "Test_Rabi_analysis_original"
   * Add test for QScale_Analysis
   * Change the way T1 value is retrieved in test_t2_analysis because now self.T1 is a dict (see above).

Images below.
Note: The plots shown for the improved versions were obtained with the changes proposed in this PR and in the Enhancement/analysis v1 plotting PR. This PR does not contain the changes form the latter PR, so the plots will look bad and the text boxes below the plots will not fit inside PNG.

@AdriaanRol 

Rabi 
2 cal_points
goo_plot

0 cal_points
good_plot

Ramsey
1 artificial detuning, for_ef==True, 6 cal_points, last_ge_pulse==True
ramsey_2nd_exc_qb2_ramsey_fit

2 artificial detunings, for_ef==False, 4 cal_points (not shown)
good_plot

QScale
qscale_qb6_qscale_fit

T1
t1_qb1_fit

Ramsey - 2 artificial detunings - without plotting improvements example
bad_plot


pycqed/tests/test_timedomain_analysis.py
```diff
@@ -24,18 +58,18 @@ def test_Rabi_analysis(self):
                 self.assertGreaterEqual(rabi_amp, amp_low)
                 self.assertGreaterEqual(amp_high, rabi_amp)
 
-        a = ma.Rabi_Analysis(timestamp='20170607_160504')
+        a = ma.Rabi_Analysis_original(timestamp='20170607_160504')
```
AdriaanRol: I very much disapprove of deprecating a method in this manner. Existing tests should not be modified! 

The reason is that we rely on them to ensure other things do not break. 
If you do want to reimplement a method (like the Rabi analysis) for which there can be good reason, it should be made such that the original tests still pass.

Vintrix:
Hi Adriaan,

I have now dealt with the Rabi_Analysis in the following way: there is only one Rabi_Analysis class now which is now a merged version of my new Rabi analysis, and what I had originally renamed to Rabi_Analysis_original (i.e. the master version). The parameter "separate_fits" (default False), lets the user choose between using the new analysis routine (False), which fits to the normalized_data_points array, or the original routine (True), which fits I and Q data separately.

I have also reverted the Test_Rabi_analysis in test_timedoamin_analysis.py to the master version, but added an additional test function to test a new feature.

I believe the overlap between this PR and the other remaining open PRs is the init of the MeasurementAnalysis base class where I define line widths, font sizes etc. I had to include this in more than on PR because the analysis classes in those PRs use them, so they expect them to be there. 
However, I have made the MA init nicer, based on your comment from the Enhancement/analysis v1 plotting PR.

Yes, this PR is a bit big :) There was no nice way to split it up because it is all concerned with time domain analysis classes, which were a big part of my semester project. 

Thanks very much for all the helpful comments, and please let me know if there is anything else I can change.

Best

AdriaanRol:
@Vintrix I would have to look at this in more detail. In the meantime I have approved the other PR's.

AdriaanRol:
@Vintrix after merging with your other PR's there are now conflicts with the master. Could you resolve these?

Vintrix:
@AdriaanRol  I resolved the conflicts with the master.

I was glad to see that you approved the other PR's. 
Take your time with this PR, it is a bit big. Let me know if I can change anything to make this PR work for you.

Vintrix:
Hi @AdriaanRol , I just resolved a minor conflict between this PR and the master. 

Could you give me a quick update from your side about the status of evaluating this PR? Is there anything we can do on our side to make this work better for you?
Thanks!

AdriaanRol:
@Vintrix I cannot spot any issues with this PR at the moment, I'm only a bit scared that it might break something. I think the only way to find out is by testing it, I'll merge it into my project branch to test it. If I don't run into anything I'll merge it.

AdriaanRol:
@Vintrix I tested the PR. 

I think most of it contains some solid improvements. I only ran into problems with the interface change of a.T2_start now being a.T2_star['T2_star']. I think you have motivated this change so I have updated our code that depends on this. 

I have also made a few changes to the ramsey analysis in that it now also shows the frequency difference between old and new (see figure below). 
I also don't like the small default markers that you seem to like so I put the default marker at `None` so that it reverts to whatever matplotlib default one has defined. 

The changes I made on my branch can be found in 7964ba8017cb8002acc8c61c451a6a7e4aef64ec. 
I would like it if you could cherry-pick these changes on top of your branch (I don't think I can edit your fork). 

I cannot gaurantee that it won't break anything else on our setups but I think we have taken all reasonable precautions. 

ramsey_1pulse_sep_ql_ramsey_fit


tl;dr if you cherry-pick 7964ba8017cb8002acc8c61c451a6a7e4aef64ec I will approve and merge.

AdriaanRol:
@Vintrix I see some changes. Let me know when you're done and I'll approve + merge.

Vintrix:
@AdriaanRol sorry for the late reply. 

I have cherry-picked your commit 7964ba8. I have also fixed the error reported by Travis. 

If all checks pass, then this PR is ready to be merged.

PR #355, opened 2017-10-04
Fix ordering of channels in UHFQC detector function

lciorciaro:
The channels where put in an array by looping over dict.keys(), which was not guaranteed to preserve ordering. By sorting the keys we guarantee the ordering is the same as the order in which the acquisition channels were defined.

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
@@ -471,7 +471,7 @@ class UHFQC(Instrument):
 
         return data
 
-    def acquisition_initialize(self, channels=set([0, 1]), mode='rl'):
+    def acquisition_initialize(self, channels=(0, 1), mode='rl'):
         # Define the channels to use and subscribe to them
         self.acquisition_paths = []
 
--- a/pycqed/measurement/detector_functions.py
+++ b/pycqed/measurement/detector_functions.py
@@ -1275,7 +1275,7 @@ class UHFQC_input_average_detector(Hard_Detector):
         data_raw = self.UHFQC.acquisition_poll(samples=self.nr_sweep_points,
                                                arm=False, acquisition_time=0.01)
         data = np.array([data_raw[key]
-                         for key in data_raw.keys()])  # *self.scaling_factor
+                         for key in sorted(data_raw.keys())])  # *self.scaling_factor
 
         return data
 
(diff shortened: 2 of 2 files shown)


elrama-:
Remark: the issue was noticed while taking tomography data. 

Note how before fix, the two readout channels are switched in two instances: 
(the right stripes are calibration points which should be constant):
download 9

After fix, we observe stability so far:

download 8

pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
```diff
@@ -471,7 +471,7 @@ def acquisition(self, samples, acquisition_time=0.010, timeout=0,
 
         return data
 
-    def acquisition_initialize(self, channels=set([0, 1]), mode='rl'):
+    def acquisition_initialize(self, channels=(0, 1), mode='rl'):
```
AdriaanRol: :+1:  tuple should lock order

pycqed/measurement/detector_functions.py
```diff
@@ -1275,7 +1275,7 @@ def get_values(self):
         data_raw = self.UHFQC.acquisition_poll(samples=self.nr_sweep_points,
                                                arm=False, acquisition_time=0.01)
         data = np.array([data_raw[key]
-                         for key in data_raw.keys()])  # *self.scaling_factor
+                         for key in sorted(data_raw.keys())])  # *self.scaling_factor
```
AdriaanRol: This sorts the data by the alphabetical order of the keys that the UHFQC returns. 
I would expect this to introduce bugs when we want to acquire data in a different order. This could happen when e.g., trying to acquire data for qubit 5 and qubit 3 in which case `channels=(5,3)` and the sort will sort them the other way round. 

@lciorciaro you explained to me that this is not the case but if I read this I expect to introduce a new error. Can you comment on this?

brianzi: It should work. The ordering is determined in a bit of a round-about way, and could and should probably be arranged a bit clearer. This PR is just a minimum-number-of-lines hotfix of the problem.

Here is how it goes: 
UHFQC.acquisition_initialize sets UHFQC.acquisition_paths (strings) according to a list of channels given to it (which are integers), (preserving the order).

acquisition_poll returns a dictionary; the keys of that dictionary are integers, but they are not channels, but indices into acquisition_paths (and thus channels, which has the same order). Looping over these keys as sorted when interpreting the data in the detector thus arranges the returned data in the same order in which channels are passed to acquisition_initialize.

PR #368, opened 2017-10-16
VNA module

quantumCryoCoder:
Necessary module for measurements with VNA.
Ramiro and Stefano wrote this module for measurements with the VNA some time ago, but it was never merged with the master.
This module is necessary for measurements with the VNA, and it needs to be merged ASAP to guarantee VNA measurements ability to other branches.


elrama-:
@AdriaanRol Usual mergings require two checks(in my opinion):
> Working code: The code must be tested and guaranteed to be working.
> Compatibility: The code must be incorporated properly in the PycQED platform.

Regarding this two points:
> Working code: The code has been actively used in every characterization since sooner than May this year (Quick glance at onenote suggests at least 13 samples have actively used this code). Moreover, this code allows to produce very fast very high quality resonator scans (example in hand, Flo's very beautiful resonator data).
> Compatibility: This code bypasses the qubit object. While this might look like is breaking the format of our measurement platform, it actually saves the day, as qubit objects might be difficult to create meaninfully before knowing the spectrum of your device.

I still think however that a dicussion about implementing spectrum characterization within our measurement software might provide us with useful changes.

In short: I agree with merging(this works perfectly in practice), but that also some changes might come in the future.

PR #371, opened 2017-10-19
Single shot readout analysis v2 

### LINKED ISSUE #370: SSRO not reliable
When measuring for the recent month report with @NielsBultink we found some problems with the current SSRO analysis. 

- If there is only a single channel of data (such as when doing optimized weights) the rotation method can cause the analysis to fail even though it is not required. 
- The plots that are generated are internally inconsistent (units shown) making it hard to debug. 
- The fitting to extract assignment fidelity does not always converge and as such sometimes crashes. 

These problems can be traced back to the ad-hoc nature of the development of the SSRO analysis having causes spaghetti code. It is increasingly relevant as we rely on it to extract thresholds for digitized readout. 

I will work on resolving this issue.

AdriaanRol:

Fixes #370 .

- ma_v2 cleaned up version of single shot readout analysis 
- extracts  
     - raw assignment fidelity 
     - fitted assignement fidelity 
     - fitted discrimination fidelity 
     - corresponding thresholds 
     - residual excitation and measurement induced relaxation 
- has options to 
     - disable figure creation (greatly speeds up analysis) 
     - plot on a log scale 

- Currently being tested in experiment (proj/m18 branch)  
- Note that the fits are constrained (mu and sigma of gaussian peaks) to prevent misfitting on the noise 
- it has tests 

image

There are some small points of improvement that I want to add before merging. 
- [x] Extract SNR (and display it) 
- [ ] facilitate easier reloading of the measurement analysis module 

I also want to test it's robustness the next couple of days to make sure it really is as robust as I claim it to be. 

This PR does not address several other issues 
- extending this analysis to mux RO (though the ingredients are there) 
- analysis for a 2D histogram (IQ shots). 

@NielsBultink I've selected you as a reviewer as you are mr measurement.


AdriaanRol:
@MalaySingh , Since you are working on tomo for measurement_analysis_v2, could you take a look at reviewing this?

AdriaanRol:
@NielsBultink I added SNR to the new SSRO analysis
1d_histogram_20171016_135112

PR #382, opened 2017-10-26
Enh/measurement analysis

AdriaanRol:

- AutoPEP 8 formatting of measurement_analysis.py
- Made plotting options configurable using `kw.get(key, default_value)`
- Changed default savefig to save with a transparant background 

For presentations it can be useful to have by default a transparant background as well as the ability to change the label color (e.g. when using a dark presentation style). 

The screenshot below (grey is how mac preview shows transparancy) shows an example of such a figure. 

screenshot 2017-10-26 12 01 47

@antsr @Vintrix I have asked you to review this as you have recently made plotting improvements to MA. I suspect this is not too controversial but let me know what you guys think.

### DIFF
--- a/pycqed/analysis/measurement_analysis.py
+++ b/pycqed/analysis/measurement_analysis.py
@@ -60,19 +60,19 @@ class MeasurementAnalysis(object):
         self.fit_results = []
         self.cmap_chosen = cmap_chosen
         self.no_of_columns = no_of_columns
-        self.dpi = 600                  #dpi for plots
-        self.axes_line_width=0.5        #lw of axes and text boxes
-        self.font_size = 11             #font sizes
-        self.tick_length = 4            #tick lengths
-        self.tick_width = 0.5           #tick line widths
-        self.marker_size=3              #marker size for data points
-        self.line_width=2               #line widths connecting data points
-        self.marker_size_special=8      #marker size for special points like
-                                        #peak freq., Rabi pi and pi/2 amplitudes
+        self.dpi = 600  # dpi for plots
+        self.axes_line_width = 0.5  # lw of axes and text boxes
+        self.font_size = 11  # font sizes
+        self.tick_length = 4  # tick lengths
+        self.tick_width = 0.5  # tick line widths
+        self.marker_size = 3  # marker size for data points
+        self.line_width = 2  # line widths connecting data points
+        self.marker_size_special = 8  # marker size for special points like
+        # peak freq., Rabi pi and pi/2 amplitudes
         self.box_props = dict(boxstyle='Square', facecolor='white',
                               alpha=0.8, lw=self.axes_line_width)
-        self.qb_name = qb_name          #for retrieving values of qubit
-                                        #parameters from data file
+        self.qb_name = qb_name  # for retrieving values of qubit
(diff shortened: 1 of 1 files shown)


pycqed/analysis/measurement_analysis.py
```diff
@@ -60,19 +60,34 @@ def __init__(self, TwoD=False, folder=None, auto=True,
...
-                                        #parameters from data file
+        self.qb_name = qb_name  # for retrieving values of qubit
+        # parameters from data file
+
+        self.tick_color = kw.get('tick_color', 'k')
+        # tick label color get's updated in savefig
+        self.tick_labelcolor = kw.get('tick_labelcolor', 'k')
+        self.axes_labelcolor = kw.get('axes_labelcolor', 'k')
+
+        params = {"ytick.color": self.tick_color,
```
antsr: What is the reason for updating the rcParams only for the color options? It would be nice if we could do it in the same way for all the style parameters.

AdriaanRol: @antsr , the reason is twofold. The main reason is that I did not have time to do this for all the options that were added earlier. The second reason is that this is not straightforward as not all properties we want to change have a one-to-one correspondance in the rcParams. An example is the tick_labelcolor, which I set by hand in the save fig. The reason is that the tick_color affects both the ticks and ticklables. For the example above we want the ticks to be black (they are inside the plot on a white background) and the ticklabels to be white (outside of the plot). 

Also @antsr if you leave comments could you (whenever possible) hit either the approve or the request changes checkbox? That way I can merge or know that I would have to change something (or convince you I don't). 

Best,

AdriaanRol: Also adding this for the title would be nice but I don't know what option I need. If you find it, let me know.

pycqed/analysis/measurement_analysis.py
```diff
@@ -60,19 +60,34 @@ def __init__(self, TwoD=False, folder=None, auto=True,
...
-        self.axes_line_width=0.5        #lw of axes and text boxes
-        self.font_size = 11             #font sizes
-        self.tick_length = 4            #tick lengths
-        self.tick_width = 0.5           #tick line widths
-        self.marker_size=3              #marker size for data points
-        self.line_width=2               #line widths connecting data points
-        self.marker_size_special=8      #marker size for special points like
-                                        #peak freq., Rabi pi and pi/2 amplitudes
+        self.dpi = kw.get('dpi', 300)  # dpi for plots
+        # 300 dpi is a good balance between resolution and speed of analysis
```
Vintrix: Writing these parameters with kw instead of hardcoding them is fixed in PR #347.

pycqed/analysis/measurement_analysis.py
antsr: Ok I see the point.

pycqed/analysis/measurement_analysis.py
AdriaanRol: @Vintrix I looked at your PR, I think putting them in a separate method is nicer (as in your PR).  I suggest we include that improvement when going over your PR (probably somewhere next week).

PR #396, opened 2017-11-06
Feature/qutech vsm driver

bergmansj:
Please use the following template for a pull request. (I will try, this is a first for me.)

Adds a qcodes driver for the QuTech VSM.
This pull request was requested by @duijedeurloo , but should probably be reviewed by
@AdriaanRol or @wvlothuizen 

The conditions below are probably not met because I wouldn't know how to perform
all these tests. Probably performing base SCPI testing of the VSM will be more effective.

- travis test suite passes

- follow the PEP8 style guide 
- have tests for the code


pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+
+        self.connect_message()
+
+    def _recv(self):
+        """ Overwrites base IP recv command to ensuring read till EOM
+            Again, this should be handled properly in a base class.
+        """
+        return self._socket.makefile().readline().rstrip()
+
+    def connect_message(self, idn_param='IDN', begin_time=None):
```
AdriaanRol: I think one should not overwrite a method like this for the reason stated. I agree with the general point but I would like to prevent any divergence form the QCoDeS base. As far as I know a proper logging solution is on the list for QCoDeS core.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+            setter = 'POWERSUPPLY:{p}'.format(p=power_supply.upper()) + ' {}'
+            # Disable setting the front power supply
+            if power_supply == 'front':
+                setter = None
+            self.add_parameter(name, get_cmd=getter, set_cmd=setter,
+                               vals=vals.OnOff())
+
+    def add_temperature_parameters(self):
+        # Command class temperature
+        self.add_parameter('setpoint',
```
AdriaanRol: It would be nice if this parameter has a unit I expect 'K' or "fancy symbol C"

AdriaanRol: Also a docstring would be nice (can be added by docstring="") , I don't know the difference between the setpoint and average temperature.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                           get_cmd='TEMPERATURE:SETPOINT?',
+                           set_cmd='TEMPERATURE:SETPOINT {}',
+                           vals=vals.Numbers(min_value=10.0, max_value=55.0))
+        self.add_parameter('temperature_avg',
+                           get_cmd='TEMPERATURE?',
```
AdriaanRol: unit+ docstring

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                           vals=vals.Numbers())
+
+        for mod in self.modules:
+            mod_name = 'mod{m}'.format(m=mod)
+            mod_scpi = 'MODULE{m}'.format(m=mod)
+            self.add_parameter(mod_name + '_temperature_avg',
```
AdriaanRol: Even more temperatures here. 
Again some docstring that tells me which one it is would be useful.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                self.add_parameter(mod_ch_name + '_temperature',
+                                   get_cmd='TEMPERATURE:'+mod_ch_scpi+'?',
+                                   vals=vals.Numbers())
+
+    def add_marker_parameters(self):
+        # Command class marker
+        self.add_parameter('marker_source',
+                           set_cmd='MARKER:SOURCE {}',
+                           vals=vals.Enum('int', 'ext'))
+        self.add_parameter('marker_state',
```
AdriaanRol: The difference between marker state and marker_source is unclear to me. Could you add a docstring?

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                           set_cmd='MARKER:STATE {}',
+                           vals=vals.OnOff())
+
+        for mod in self.modules:
```
AdriaanRol: You seem to both add the parameter once and after that add it on a per module basis. First what is the difference between the two and why do you add it twice?

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                self.add_parameter(mod_ch_name + '_marker_source',
+                                   get_cmd='MARKER:'+mod_ch_scpi+':SOURCE?',
+                                   vals=vals.Enum('int', 'ext'))
+
+    def add_qubit_parameters(self):
+        # Command class qubit
+        for channel in self.channels:
+            ch_name = '_ch{c}'.format(c=channel)
+            ch_scpi = ':CHANNEL{c}'.format(c=channel)
+            self.add_parameter('qubit' + ch_name + '_description',
```
AdriaanRol: what is qubit description used for? Is this used for the monitor/display? Also please add a docstring.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                               get_cmd='QUBIT'+ch_scpi+':DESCRIPTION?',
+                               set_cmd='QUBIT'+ch_scpi+':DESCRIPTION {}',
+                               vals=vals.Strings())
+            self.add_parameter('qubit' + ch_name + '_frequency',
```
AdriaanRol: What units is this in? It should be in Hz (we do everything in SI). If the device thinks e.g. in GHz then please add a conversion. 

Also could you add a docstring on what this is used for? I think it relates to what calibration data is used.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                               get_cmd='QUBIT'+ch_scpi+':FREQUENCY?',
+                               set_cmd='QUBIT'+ch_scpi+':FREQUENCY {}',
+                               vals=vals.Numbers())
+            self.add_parameter('qubit' + ch_name + '_led_color',
```
AdriaanRol: what kind of strings are accepted here? basic color names? hex values? Please add a docstring.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                mod_ch_name = '_mod{m}_ch{c}'.format(m=mod, c=channel)
+                mod_ch_scpi = ':MODULE{m}:CHANNEL{c}'.format(m=mod, c=channel)
+                self.add_parameter('qubit' + mod_ch_name,
+                                   get_cmd='QUBIT'+mod_ch_scpi+'?',
+                                   set_cmd='QUBIT'+mod_ch_scpi+' {}',
+                                   vals=vals.OnOff())
+
+    def add_parameters(self):
+        # VSM properties
+
```
AdriaanRol: I like the grouping of add parameters here.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+        self.add_marker_parameters()
+        self.add_qubit_parameters()
+
+        # Command class calibration
+        for pulse in ('GAUSSIAN', 'DERIVATIVE'):
+            for ap in ('ATTENUATION', 'PHASE'):
+                # All channels and modules (no getter)
+                var_name = '_{p}_{a}_raw'.format(p=pulse.lower(), a=ap.lower())
+                var_scpi = ':{p}:{a}:RAW'.format(p=pulse, a=ap)
+                self.add_parameter('calibration' + var_name,
```
AdriaanRol: What is a calibration "Int"? and what does it do? Please add a docstring (and if it refers the state it should be a Bool).

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+                                           get_cmd='CALIBRATION' +
+                                                   mod_ch_scpi +
+                                                   var_scpi + '?',
+                                           set_cmd='CALIBRATION' +
+                                                   mod_ch_scpi +
+                                                   var_scpi + ' {}',
+                                           vals=vals.Ints())
+
+
+if __name__ == '__main__':
```
AdriaanRol: I guess the __name__=='__main__': is there for testing purposes. We generally use drivers as a module and as such this bottom part should be removed to keep the code clean. 
Adding an example to the examples folder with this code snippet would be useful but not required.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
```diff
@@ -0,0 +1,267 @@
...
+        self.add_power_supply_parameters()
+        self.add_parameter('mbbc_state', get_cmd='MBBC?',
+                           vals=vals.Enum('connected', 'disconnected'))
+        self.add_temperature_parameters()
+        self.add_marker_parameters()
+        self.add_qubit_parameters()
+
+        # Command class calibration
+        for pulse in ('GAUSSIAN', 'DERIVATIVE'):
+            for ap in ('ATTENUATION', 'PHASE'):
```
AdriaanRol: I can see the calibration paramter but I can find nowhere where we can set the channel specific attenuation/phases. Also (once these are added) what units are these in? dB? raw dac values?

AdriaanRol:
Hi @bergmansj my main review comments can be summarised in the following points: 
- There are no docstrings and units making it very hard to understand how things work. 
- You seem to be adding every parameter once globally and then again for each module. This seems like a bug unless I misunderstand. 
- The parameters to set the channel specific Attenuation and Phase do not seem to be added. 

I would greatly appreciate it if you could resolve these issues.

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
bergmansj: Fixed this by inheriting from SCPI (just like the AWG driver)

pycqed/instrument_drivers/physical_instruments/QuTech_VSM_Module.py
bergmansj: Fancy symbol it is :)

bergmansj:
Changes:

- Inherit from SCPI, like the AWG driver.
- Added parameter docstrings and extended module documentation.
- Where applicable added units.
- Removed parameters from the driver that have no prio yet.
- Removed commands for temperature setpoints (of the heating PID controllers)
  because we probably don't want to allow for changing those.
- No validators for getters. (why are there validators on the driver side at all?)
- Removed `__main__` block from module.
- Calibration commands for each qubit separately, to keep things simple. This
  can be extended later when requested.
- During development, internal VSM logs can be viewed by pointing a
  webbrowser to port 5000 of the VSM, example: http://192.168.0.10:5000/
  For normal use, this will probably be restricted/disabled.
- Some code-yoga to adhere to PEP8 :)

bergmansj:
Sorry for not responding earlier this week. I processed all your comments/suggestions. Please check them out.

NielsBultink:
Here, my first impressions.
several issues with the driver and functioning of the VSM as it is now:

**blocking:**
- stability:
when reading out all commands: for instance VSM.print_human_readable_snapshot() or station.add_component(VSM), the device often loses communication. Only way to solve this is power cycling. 

- individual marker_source settings lead to a crash
image

**not blocking but major drawback:**
- all integer and float get commands come back as strings: temperatures, attenuations, phases

**not implemented:**
- as it seems,orthogonalization of phase and attenuation settings using the on-board calibration files. Also, rescaling of the amplitudes a float between 0.x and 1 is not implemented

bergmansj:
Thanks for looking into the driver code. 

- stability:
when reading out all commands: for instance VSM.print_human_readable_snapshot() or station.add_component(VSM), the device often loses communication. Only way to solve this is power cycling. 

_These are probably methods added by the Qcodes magic. I didn't put them into the driver Too advanced for the VSM at this point maybe? what SCPI commands are involved?_ 

- individual marker_source settings lead to a crash

_The marker _source_ can only be set for a complete module, not er channel._

- all integer and float get commands come back as strings: temperatures, attenuations, phases

_Probably a result of me misunderstanding the Qcodes driver magic that is needed. The SCPI protocol is string based, so everything has to be transported as strings. Feel free to correct the driver code for this._

- as it seems,orthogonalization of phase and attenuation settings using the on-board calibration files. Also, rescaling of the amplitudes a float between 0.x and 1 is not implemented

_What do you mean by this? The calibration files are not implemented yet, because it basically is syntactic 'linarization' over the control of the attenuation and phase DACs. I even removed the preparatory code from the driver to keep things better testable._

AdriaanRol:
@bergmansj 
> Thanks for looking into the driver code.

>> stability:
>> when reading out all commands: for instance VSM.print_human_readable_snapshot() or station.add_component(VSM), the device often loses communication. Only way to solve this is power cycling.

> These are probably methods added by the Qcodes magic. I didn't put them into the driver Too advanced for the VSM at this point maybe? what SCPI commands are involved?

The magic here is almost non existent, the station.add_component reads out all parameter (calls `.get` on every parameter) before adding it to a list of components. Something as basic as reading out parameters should not lead to a crash. The SCPI commands involved are whatever get commands added to the driver. 

>> individual marker_source settings lead to a crash

> The marker source can only be set for a complete module, not er channel.

If an action is not supported in the hardware, the driver should not allow us to do this. More importantly, if we try to do something that is not allowed it should raise an exception, not crash the VSM. 

>> all integer and float get commands come back as strings: temperatures, attenuations, phases

> Probably a result of me misunderstanding the Qcodes driver magic that is needed. The SCPI protocol is string based, so everything has to be transported as strings. Feel free to correct the driver code for this.

It is as simple as adding a `get_parser=float` (in case of a float) as an argument to the add_parameter. I have no problem adding this but would prefer not to touch the driver until it has been properly tested. 


>>  as it seems,orthogonalization of phase and attenuation settings using the on-board calibration files. Also, rescaling of the amplitudes a float between 0.x and 1 is not implemented

> What do you mean by this? The calibration files are not implemented yet, because it basically is syntactic 'linarization' over the control of the attenuation and phase DACs. I even removed the preparatory code from the driver to keep things better testable.

We were under the impression that the VSM was completely done and that the driver would support the orthonormalized/linear mode. If this is not the case I guess we should discuss with @duijedeurloo and Jeremy correct? 

Best, 

Niels and Adriaan

NielsBultink:
As discussed with Jeroen, here a simple example code to make the unit crash:
In this, example I have made to attempts to avoid crashes:
- add a sleep time between each setting, still experimenting with making this longer
- add parameter assertion after each parameter setting

image

crashing throws the following error:
image

bergmansj:
@NielsBultink I made some fixes so that the code as pasted above (next time copy-pastable text please) does no longer crash the message queue.

NielsBultink:
ah, interesting. Is this already pushed to the feature branch?

2017-11-21 16:31 GMT+01:00 Jeroen Bergmans <notifications@github.com>:

AdriaanRol:
image

Looks like it.

NielsBultink:
not to me

AdriaanRol:
@NielsBultink have all the changes made while testing today been committed to this branch? 
If so we can merge.

PR #403, opened 2017-11-10
Uhfqc awg 8 test

gtaifu:

Fixes #issuenumber.

- Updated the assembler from V.02.0 to v1.1.1. A lot of improvement has been done, including:
  - Better error messages
  - Enabling dumping current configuration
  - Explicit getVersion() function
  - More...
- Improved the instruction uploading function in the CCL driver. So a binary file is no longer generated.

@AdriaanRol


pycqed/instrument_drivers/physical_instruments/QuTech_CCL.py
```diff
@@ -78,22 +74,14 @@ def _initialize_insn_microcode_parsers(self):
...
-        self.run(0),
-        self.enable(0)
-
-    def start(self):
-        self.enable(1)
-        self.run(1),
-
     def add_parameter(self, name,
-                      parameter_class=Parameter, **kwargs):
+                      parameter_class=StandardParameter, **kwargs):
```
AdriaanRol: Can you undo this replacement? It has to do with a recent update to QCoDeS, StandardParameter is now depracated. If you pull the latest version of QCoDeS from DiCarloLab-Delft/QCoDes/master you should have the right version.

AdriaanRol:
Hi @gtaifu thanks for the PR. 

There are a few things that should be fixed before I will merge this. 
- The tests are failing because it complains there is no pyQISA assembler (I added a screenshot) while the same tests in the Proj/NovDemo branch are passing. I see no changes to the tests so it must be something you introduced. 

https://www.dropbox.com/s/h5trogqbw8rmzzy/Screenshot%202017-11-10%2010.57.47.png?dl=0
- With the exception of a single comment I assume that the changes to the CCL driver are all OK (no way for me to test it). 
- There are changes to the UFHQC driver, is this intended? If not, please remove from the PR. 
- Could you undo the changes to the `qisa_opcodes.qmap`? I'm using the one that OpenQL created based on my config. 
- I assume all changes to the assembler to be OK

gtaifu:
@AdriaanRol 
- I have reverted the "StandardParameter" to "Parameter".
- The assembler cannot be found because the build directory is completely ignored so the server does not have the one that has been built.  So the online test should not be able to find the assembler. 
- I changed the UHFQC file because I am always using it. However, this change might disturb your code. I will further look into it.

Regarding the .qmap file. It is not a nice for me to revert my version back to the original version since you are using the one generated from OpenQL. Since everybody is using pycqed for different purposes so that the configuration might be different. And for my test, I do require a fixed .qmap file that might be different to yours.  We require another solution that can solve this problem from the root.

I am discussing with Vincent to make the assembler importable like OpenQL. This should in principle enable different persons to use different .qmap files.

Due to the above reasons, I do not think this pull request is good enough. I will make another branch and a new PR to update the CCL driver.

AdriaanRol:
@gtaifu I assume this is superceded by #409 ? If so can I close this PR?

AdriaanRol:
Superceded by #409 closing this PR (@gtaifu let me know if this is incorrect).

PR #409, opened 2017-11-19
Hotfix/update ccl driver

gtaifu:
Updated the CCL driver and assembler.


pycqed/instrument_drivers/physical_instruments/_CCL/qisa-as/qisa_version.h
```diff
@@ -1,8 +1,8 @@
-#define QISA_MAJOR_VERSION 0
+#define QISA_MAJOR_VERSION 1
 
-#define QISA_MINOR_VERSION 2
+#define QISA_MINOR_VERSION 1
```
AdriaanRol: Should it not be 1.0.0 (rather than 1.1.1?)

PR #411, opened 2017-11-20
Enhancement/#410/execute helpers arg change

kel85uk:
@AdriaanRol I needed to change the execute helpers interface to support having tqisa/tasm and also add a qasm file url argument because the Spin side needs the qasm file to differentiate between the qubit measurements. I don't expect it to break the system.


kel85uk:
Closing this as there seems to be too much junk

Issue #413, opened 2017-11-20
Resolving x axis 'None' for QX simulator in Quantum Infinity


kel85uk:
Resolve issue [128](https://github.com/DiCarloLab-Delft/Quantum_Infinity/issues/128) of QI

PR #414, opened 2017-11-20
Fixes the x-axis for QX simulator in Quantum Infinity

kel85uk:
Fixes issue #413 by just adding the parameter_name and unit for the qx simulator in the simulation_helper file.


kel85uk:
@AdriaanRol Could you have a quick check and see if this solves the x-axis label problem with quantum infinity? If so, please merge the PR.

PR #428, opened 2018-01-11
Periodic merge into the master to prevent wildly diverging branches. 

AdriaanRol:
This will not be a clean pull request, however it does contain a lot of updates, mostly related to QuSurf  hardware and qubit objects. 

Branches related to this are already being used by multiple people. Let's merge into the master so we can start a bit cleaner with the new year.


elrama-:
I sorted out the conflict that was there (just a different label assigned in a Ramsey mmt call).

PR #436, opened 2018-02-07
Added function detector list

elrama-:
- Added a list detector

@AdriaanRol This list detector has been extensively used in LaVespa, LaDucati, LaFerrari and simulations but for some reason it keeps getting lost in the git universe.


AdriaanRol:
As discussed with @elrama- , this method is removed because it is deprecated. Testing indicates that the old use cas can indeed be implemented using the new "Function_Detector". Depending on further feedback we will close this PR.

PR #449, opened 2018-04-06
Update Fridge Monitor for Bluefors fridge

brianzi:
Update the fridge monitor to know about Bluefors fridge.  Also cleaned the fridgemon code in a few places. Tested, still works with all other fridges.

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/Fridge_monitor.py
+++ b/pycqed/instrument_drivers/physical_instruments/Fridge_monitor.py
@@ -172,8 +172,7 @@ class Fridge_Monitor(Instrument):
         if time_since_update > (self.update_interval()):
             self.temp_dict = {}
             self.press_dict = {}
-            # try:
-            if 1:
+            try:
                 s = urlopen(self.url, timeout=5)
                 source = s.read().decode()
                 self.source = source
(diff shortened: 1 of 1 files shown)


pycqed/instrument_drivers/physical_instruments/Fridge_monitor.py
```diff
@@ -112,13 +171,16 @@ def _update_monitor(self):
         time_since_update = time() - self._last_temp_update
         if time_since_update > (self.update_interval()):
             self.temp_dict = {}
-            try:
+            self.press_dict = {}
+            # try:
+            if 1:
```
AdriaanRol: @brianzi I see you have replaced the try/except statement with an always True mode. The try/except is there for the experiments to keep running when the fridge mon website gives an error for some reason. Could you explain why you changed this? (I assume it was for debugging). 

Maybe we should add a comment as to why the try/except is there.

pycqed/instrument_drivers/physical_instruments/Fridge_monitor.py
```diff
@@ -127,19 +189,19 @@ def _update_monitor(self):
...
                     elem[1]) for elem in temperaturegroups}
                 pressuregroups = re.findall(
-                    r'<br>(P_[\w_]+(?: \(P\))?) = ([\S\.]+)', str(source))
+                    r'<br>(P_[\w_.]+(?: \(P\))?) = ([\S\.]+)', source)
                 self.press_dict = {elem[0]: float(
                     elem[1]) for elem in pressuregroups}
 
-            except Exception as e:
-                logging.warning(e)
+            # except Exception as e:
```
AdriaanRol: See comment above, if we decide against try/except, please remove it otherwise reimplement the try/except.

brianzi:
My bad, I indeed removed the try/except clause for debugging and forgot to restore it. I changed it back.

PR #455, opened 2018-04-11
Enh/m av2

tstavenga:

Fixes #issuenumber.
Push my changes to the new analysis
It is not finished but i think it is good to have a common starting point

- 
- 
-

@mention AdriaanRol

- travis test suite passes

- follow the PEP8 style guide 
- have tests for the code


AdriaanRol:
@tstavenga I see that the last change you added broke the tests, might be worth checking out why that is. 

I also see that this PR essentially contains two big changes. 
- updates to base analysis
- spectroscopy analysis file

I would propose the following
- we hand merge the base analysis that we use with the one you use (better than having the `commit fight` that Rene joked about). 
- fix the tests that are broken (you break it, you fix it) 
- we set up a basic test for the spectroscopy analysis (so that if I break it I will have to fix it)
    - it can serve as an example how to use it (e.g., how to call them, what kind of data it expects
    - it does not have to be very rigorous, just running all the code would already catch 90% of the bugs. 
- after that I go over the PR in detail. (I see some details like commented out code and such). 

Let me know if I can be of any help.

pycqed/analysis_v2/base_analysis.py
```diff
@@ -461,38 +488,31 @@ def save_fit_results(self):
         """
         Saves the fit results
         """
-
```
Aypac: This is a bugfix and needs to be merged over

Aypac:
This is not working. Created a new branch #458 to realize this request.

PR #459, opened 2018-04-16
Feature coherence analysis

Aypac:
Implements the analysis of Coherence times vs Current measurements


Aypac:
Merge error???!

AdriaanRol:
@Aypac I could not see if this analysis has a test. If it doesn't we should add one as it is quite easy to add and we found it really helps preventing errros. #whobrokeitfixesit

Aypac:
replaced by #464

PR #473, opened 2018-05-18
AWG8 Real-time bounce correction simulation functions and test file

ysalatheZI:
- Added functions multipath_bounce_correction and ideal_inverted_fir_kernel
- Wrote at test file for those new functions

@AdriaanRol


AdriaanRol:
Hi Yves, 

Thanks for the quick PR! And thanks for addressing the codacy concerns. As I am fixing a bounce error at the moment, I think I will be able to test your updates today (or this weekend). 

There are two small changes that I would like to request. 
- Add a test. 
    What I mean by this is the following, there is a file `pycqed/tests/test_kernel_distortions.py`, this contains tests for the "old" style kernels. I would like it if you create a similar test for the new kernel you have implemented. This could be a simple as just calling the function with some known arguments and seeing that you get the expected output.  

- Add a function that deals with this problem in natural units. 
On top of the file `kernel_functions_ZI.py` there are two functions, `bias_tee_correction` and `exponential_decay_correction` that perform the filtering in natural units. I would like it if you add a function that wraps around your existing function that looks similar to what I put below: 

```python 

def bounce_correction(ysig, tau:float, amp: float,
                      sampling_rate:float = 1):
    """
    Corrects for a bounce

    Args:
        ysig: the signal to be predistorted
        tau: the time at which the bounce occurs
        amp: the amplitude of the bounce correction
        sampling_rate: the sampling rate of the signal
    returns:
        filtered_signal : the signal corrected for the bounce
    """
   # here is some transformation from the input params to your named params 
   return multipath_bounce_correction(sig, delay, amp, paths = 8, bufsize = 128)

```
Could you let me know if that is OK? Again, thanks for the quick turnaround.

ysalatheZI:
Hi Adriaan,
I added the function "first_order_bounce_corr", which is a wrapper to the multipath function.
The unit tests are implemented in pycqed/tests/test_kernel_distortions_ZI.py.

Please let me know whether you are satisfied with these additions.
Best,
Yves

ysalatheZI:
Hi Adriaan,

one interesting insight I had from my bounce correction tests today, was that considering only one reflection (what I call "first-order"), can lead to errors as big as 1%, when the bounce amplitude is 10% of the pulse amplitude. This can be seen when running the script examples/AWG_examples/bounce_correction_test.py, which is part of this PR.

So it is important to assess whether the "first-order" correction is good enough for you or whether you are going to need a more accurate solution.

So please let me know your thoughts.

Have a nice weekend!

Best,

Yves

PR #488, opened 2018-06-22
Precomp

ysalatheZI:
- Added new function 'run_realtime_predistortion' to the ZI_HDAWG8 device driver.
- Added new nodes for controlling the AWG8 trigger outputs to node_doc_HDAWG8.json.
- Added a Jupyter notebook with information on how to use the function run_realtime_predistortion to the AWG8_examples.

@AdriaanRol : The added functionality should make it easier to use the real-time pre-distortion with the HDAWG V1 devices. Let's discuss the further steps.


AdriaanRol:
@ysalatheZI I will take a detailed look at this when the new v1 devices arrive in Delft next week and I can test the feature. 

However, from a quick scan I have a comment. I see that the `run_realtime_predistortion` method requires passing the arguments for the predistortion filters to be used. This goes against the core idea of storing settings like these in parameters in order to ensure that the settings can be reconstructed after an experiment. 

I would suggest implementing it in a slightly different way. 
1. Add parameters to the HDAWG driver to contain the filter coefficients. Most likely this equates exposing the parameters that are set within the method directly to the ZIPython `listNodesJSON` method. If you call the `create_parameter_file` method when you have created the instance they should show up in the parameter file and become part of the driver automatically. 
2. Replace the `run_realtime_distortion` method with a parameter that enables/disables the distortions on a per channel basis. 

If there is a specific order in which everything needs to be set for it to work properly then it makes sense to create a distinction between a private version of these parameters and a public version. The public version is then what is logged (could be done by adding `ManualParameter`s) and the private is part of the ZIPython API that is not exposed. 

Best, 

Adriaan

AdriaanRol:
Closing pull request as changes have been merged into the `develop_awgv1_devices` branch with some modifications.

PR #497, opened 2018-08-28
Cz flux arc fix

AdriaanRol:
Radical changes to the CZ flux pulses, made with @FKMalina . 

Please do not merge, this is mostly there so I have access to a convenient diff with develop.


AdriaanRol:
Everything is implemented, @FKMalina is using it now in experiment. 
I will write some more tests to ensure everything is working as expected and plot some figures below of the plotting methods (that still need to be updated).

AdriaanRol:
Images generated by the flux lutman on demand now. 

image
image

fbattistel:
I've realized that the bug was not affecting the final result because we take the ratio t/scale afterwards

AdriaanRol:
Hi @fbattistel , I'm assuming you are OK with the changes but could you (for the sake of protocol) hit the approve button? 

You can find it in the review changes button in the "files changed" tab. 
screenshot 2018-08-30 13 40 57

Please do not merge yet, this is typically done by whoever opened the pull request (PR).

Issue #508, opened 2018-10-02
Analysis for pulsed Ramsey experiments 


AdriaanRol:
Should link up with other noise characterization analysis. 
Use the individual analysis by @FKMalina for the separate files.

AdriaanRol:
Update on progress: 
There were two versions of this analysis, one made by @Aypac and one existing in notebooks that has been passed around from @quantumCryoCoder to @elrama- to @FKMalina . The second version only does not include data extraction but is a lot cleaner. 

The current status is: 
- Both variants give the same values for the derived quantities
- A lot of bugs have been squashed 
- The following (nice figures ) are now created 
- It is now a v2 analysis. 


![image](https://user-images.githubusercontent.com/6142932/46485145-69c46f00-c7fb-11e8-8a18-df7bbf782fd4.png)
![image](https://user-images.githubusercontent.com/6142932/46485152-6df08c80-c7fb-11e8-8007-7fa711dd3d61.png)
![image](https://user-images.githubusercontent.com/6142932/46485156-72b54080-c7fb-11e8-843a-1d8ac80380a0.png)
![image](https://user-images.githubusercontent.com/6142932/46485206-85c81080-c7fb-11e8-9341-22293a531abe.png)

The big problem is that I (and I'm not alone here) do not understand exactly what values are there. The code still contains a lot of magic numbers with references to papers that do not exist. (e.g., Martinis PRA 2003 (there does exist a PRB 2003)). 

I'll have to look into the physics to understand the details here, also because I want to use a modified version to understand the flux noise for the pulsed experiments.

MiguelSMoreira:
Closed due to inactivity

PR #512, opened 2018-10-08
Flux noise/coherence analysis

AdriaanRol:

Relates to #508 .

- A single v2 analysis that works from the point where data is presented in a table
- Clear comments on what is exactly analyzed 
- Should replace old code that is currently scattered in notebooks and spaghetti. 

Some things I wanted to do but have not yet been able to finish 
- completely kill the old analysis, can't do because I haven't yet done the data extraction automation
- automate this for pulsed experiments. I've done it using the same core functions but have not yet integrated this. 

I want to merge this so that @NielsBultink can start using the latest version.


pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -907,150 +1197,404 @@ def prepare_plots(self):
...
+        Qc:             coupling Q of the readout resonator
+        chi_shift       in Hz
+        path:           filepath, if provided is used for saving the plots
+
+
+
+    Input table specification:
+           Row  | Content
+        --------+--------
+            1   | dac
```
FKMalina: Explain what dac is. Current? Flux? Any of those?

pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -907,150 +1197,404 @@ def prepare_plots(self):
...
     """
-    val_string = par_name
-    val_string += ': {:.3g}'.format(lmfit_par.value)
-    if lmfit_par.stderr is not None:
-        val_string += r'$\pm$' + '{:.3g}'.format(lmfit_par.stderr)
+    dac, freq, T1, Tramsey, Techo, exclusion_mask = table
+    exclusion_mask = np.array(exclusion_mask, dtype=bool)
+
+    # Extract the dac arcs required for getting the sensitivities
+    fit_result_arch = fit_frequencies(dac, freq, dac0_guess=.1)
```
FKMalina: Fails to fit dac arc when I'm using dac that is already in flux quantum units. Needs better dac0_guess. Perhaps based on parabola fit?

pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -1,27 +1,302 @@
...
+            - fit dephasing rates
+            - determine derived quantities
+        """
+        self.fit_res = OrderedDict()
+        self.proc_data_dict = deepcopy(self.raw_data_dict)
+        pdd = self.proc_data_dict
+
+        # Extract the dac arcs required for getting the sensitivities
+        # FIXME: add a proper guess function to make the fit more robust
+        self.fit_res['dac_arc'] = fit_frequencies(pdd['dac'], pdd['freq'],
```
FKMalina: Fails to fit dac arc when I'm using dac that is already in flux quantum units. Needs better dac0_guess. Perhaps based on parabola fit?

pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -907,150 +1197,404 @@ def prepare_plots(self):
...
+        gamma_phi_echo      dephasing rate of echo experiment
+        verbose (int)       verbosity level
+    Returns:
+        fit_result_gammas  an lmfit fit_res object
+    """
+    # create a parameter set for the initial guess
     p = lmfit.Parameters()
     p.add('slope_ramsey', value=100.0, vary=True)
     p.add('slope_echo', value=100.0, vary=True)
     p.add('intercept', value=100.0, vary=True)
```
FKMalina: Should there be a lower bound of 0 on the intercept?

pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -1,27 +1,302 @@
...
+
+    Coherence_table specification:
+           Row  | Content (arrays)
+        --------+--------
+            1   | dac
+            2   | frequency (Hz)
+            3   | T1 (s)
+            4   | T2 star (s)
+            5   | T2 echo (s)
+            6   | Exclusion mask (True where data is to be excluded)
```
FKMalina: Figuring out to set to True in the exclusion mask is a huge pain. Additional helper plots would be great. Eg. one with that use the an index as a marker for the corresponding point. Just make sure dpi is large enough to make it readable. Or use vector format.

pycqed/analysis_v2/coherence_analysis.py
```diff
@@ -1,27 +1,302 @@
...
+
+    Generates 8 plots:
+        > T1, T2, Echo vs flux
+        > T1, T2, Echo vs frequency
+        > T1, T2, Echo vs flux sensitivity
+        > ratio Ramsey/Echo vs flux
+        > ratio Ramsey/Echo vs frequency
+        > ratio Ramsey/Echo vs flux sensitivity
+        > Dephasing rates Ramsey and Echo vs flux sensitivity
+        > Dac arc fit, use to assess if sensitivity calculation is correct.
```
FKMalina: Add description of where in the analysis object the relevant parameters are stored. Half of the time when I use analysis object I don't care about the plots, only about fitted parameters. Say I want to use \sqrt(A) for ramsey. I'd rather not dig through the code to figure out how to get access to that in my analysis notebook.

pycqed/analysis_v2/coherence_analysis.py
AdriaanRol: yes.

pycqed/analysis_v2/coherence_analysis.py
FKMalina: (Philosophical comment) This is one of the reasons why everyone writes their own scripts. It is less frustrating and easier to write your own script then to figure out where in the analysis object the results are stored.

pycqed/analysis_v2/coherence_analysis.py
AdriaanRol: I think that is a good addition.

pycqed/analysis_v2/coherence_analysis.py
AdriaanRol: Excellent point.

PR #514, opened 2018-10-19
Enh/configurable gateset

AdriaanRol:
Makes the gateset of the lutman configurable.


AdriaanRol:
@slavoutich , I forgot why we didn't merge this yet.

pycqed/measurement/openql_experiments/generate_qi_cfg.py
```diff
@@ -54,13 +54,13 @@ def _rotation_instruction(phi, theta):
```
slavoutich: Shouldn't generate config files be removed at all already now?

pycqed/measurement/openql_experiments/openql_helpers.py
```diff
@@ -13,6 +13,7 @@
 
 output_dir = join(dirname(__file__), 'output')
 ql.set_option('output_dir', output_dir)
```
slavoutich: Helpers are not extremely needed anymore, IMO.

pycqed/measurement/openql_experiments/generate_qi_cfg.py
AdriaanRol: No, we can remove this only once QINF is properly integrated.

pycqed/measurement/openql_experiments/openql_helpers.py
AdriaanRol: Same as above.

AdriaanRol:
This was closed manually.

PR #515, opened 2018-10-23
Add realistic error model for qubits and CZ

slavoutich:
No functional changes.


slavoutich:
Also it might be useful to add realistic readout parameters, but I don't know them.

AdriaanRol:
Rejecting as it is superseded by the configurable gate set.

PR #516, opened 2018-10-25
Bridging/nanowire

tstavenga:
merging into development to stay in sync.


ElmoreV:
Overnight, the package pyGsti (for gate set tomography) has been updated to 0.9.7, breaking backwards compatibility with previous versions. This breaks the code in this branch. Was this intentional? And if so, will there be an update to master to fix it? (This branch works on pyGsti version 0.9.4.4)

AdriaanRol:
@ElmoreV , tests have passed (Huraay!) . Take a look at the codacy (there are two minor things to fix there) and be sure to remove the files that should not be and I think we are good to merge.

wvlothuizen:
> pyGsti

I recently attended a presentation by Sandia on pyGsi (at the LogiQ/QuSurf TEM) where they explained that they broke compatibility in order to be able to move forward. They also mentioned the availability of upgrade scripts, and the fact that they are willing to assist if necessary,

AdriaanRol:
@wvlothuizen , I agree on the upgrading PyGSTi, however we still need our test builds to pass before we upgrade. So for now I have locked in our version to the latest version before their upgrade. IMO upgrading our PyGSTi code is a separate issue (one of the many).

ElmoreV:
@AdriaanRol The last commit should make the pull request ready for merging.

PR #524, opened 2018-11-30
Precomp FIR filter

ysalatheZI:
Additions for the real-time finite impulse response (FIR) filter.

- Added a simple function "build_piecewise_kernel", which generates the FIR filter kernel with pairwise-equal coefficients from a list of coefficients.
- Updated the node documentation of the HDAWG8.
- Provided an example of how to use the real-time FIR filter with PycQED.

Note: The real-time FIR filter requires LabOne 57469 or later. See the QuSurf Issue Tracker Wiki for the specifications.


AdriaanRol:
@ysalatheZI 
We have done a simple test of the realtime FIR filters. 
In this test we use a UFHQC connected directly to the output of the HDAWG to measure the effect of the FIR filter and compare this to the prediction/model. 

In the data below we can see that the prediction does **NOT** match the filter. 
The experiment that is done is the following. 
1. Measure a square pulse with no FIR predistortion
2. Generate an FIR filter and predict the effect using the example from the provided notebook. 
3. Upload the FIR filter and measure the 

fir_test

We have attached the notebook used (modified version of notebook provided in this PR) as a zip in this comment. 
For us it is essential that the model/predictions of the filter match the effect of the implemented filter. As such, we request the model to be updated to provide an accurate prediction of the effect of the filter. 

[Realtime_predistortion_test.zip](https://github.com/DiCarloLab-Delft/PycQED_py3/files/2894244/Realtime_predistortion_test.zip)

ysalatheZI:
See comments in the Issue tracker.

ysalatheZI:
I've now figured out the actual issue: for the prediction, the FIR filter was applied onto the signal measured by the UHF, which has a sample rate of 1.8 GSa/s. However, the HDAWG applies the FIR filter onto a sample stream with rate 2.4 GSa/s. Therefore, it is essential to first resample the measured signal using interpolation before applying the FIR filter:
```
t_UHF=np.arange(len(values_ref))/1.8e9 - BUFFER + UHF_TRIGGER_LATENCY
t_HDAWG = np.arange(min(t_UHF), max(t_UHF), 1/2.4e9)

values_ref_HDAWG = np.interp(t_HDAWG, t_UHF, values_ref)
pred_response_HDAWG = signal.lfilter(fir_kernel, 1.0, values_ref_HDAWG)
```
When doing that, the prediction gets much more accurate:
fir_test

Even more so for the Hanning-Window:
fir_test_hann

Again I enclosed the new Jupyter notebook for reference:
[Realtime_predistortion_test.zip](https://github.com/DiCarloLab-Delft/QuSurf-IssueTracker/files/2899360/Realtime_predistortion_test.zip)

Does that solve your problem?

AdriaanRol:
Closing this PR as all changes have already been incorporated in the develop branch by hand.

PR #538, opened 2019-04-15
Fix CI builds

slavoutich:
The problem was missing libxkbcommon-x11.so.0 library in the image, that lead to the fact that virtual display failed to initialize. Fixed in Dockerfile.


slavoutich:
I am running CI in manual mode now, if it passes -- we can merge.

PR #543, opened 2019-04-29
Bug/502 analysis robust failing fits

### LINKED ISSUE #502: Analysis not robust against failing fits
When a fit does not converge, analysis typically "fail". 

However, there are two problems with this
- The failure occurs when trying to use the parameters standard error property in text parsing and not in a clear way. 
- This yields unclear error messages 

It would be preferred if the analysis gives a better indication of what is going on. 

Often the analysis itself is still useful (i.e., plotting the data etc) only the fit is unreliable. In those cases no exception should be raised.

AdriaanRol:
@slavoutich , Made this over the weekend. 
Should address most instances of failing fits causing analysis to break. 
Fixes #502 .

- Adds a  format_value_string that formats an lmfit parameter in a unit aware way 
- Adds a "safe_formatter" that formats a string and replaces `None` and invalid values by `~~` and `!!`
- Replaced printing in some (but not all analysis). 

This should probably be done in all analysis but I did not have time for this. I suggest we add this type of formatting whenever we encounter a new analysis that fails because of this reason.

### DIFF
--- a/pycqed/analysis_v2/readout_analysis.py
+++ b/pycqed/analysis_v2/readout_analysis.py
@@ -23,7 +23,7 @@ from pycqed.analysis.tools.plotting import SI_val_to_msg_str
 import pycqed.analysis.tools.data_manipulation as dm_tools
 from pycqed.analysis.tools.plotting import set_xlabel, set_ylabel
 from pycqed.utilities.general import int2base
-
+from pycqed.utilities.general import format_value_string
 
 class Singleshot_Readout_Analysis(ba.BaseDataAnalysis):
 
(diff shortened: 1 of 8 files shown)


pycqed/analysis_v2/readout_analysis.py
```diff
@@ -679,24 +679,29 @@ def prepare_plots(self):
...
                 a_sp = bv['A_spurious']
                 fit_text += '\n\nSpurious Excitations:'
-                fit_text += '\n$p(e|0) = {:.3f}$'.format(a_sp.value)
-                if self.options_dict.get('fixed_p01', True) == True:
-                    fit_text += '$\\pm{:.3f}$'.format(a_sp.stderr)
-                else:
-                    fit_text += ' (fixed)'
+
+                fit_text += format_value_string('\n$p(e|0)$', lmfit_par=a_sp)
+                # fit_text += '\n$p(e|0) = {:.3f}$'.format(a_sp.value)
```
slavoutich: Can you remove these comments?

Issue #31, opened 2016-09-30
QASM to Latex visualization 


AdriaanRol:
Adapt the Nielsen and Chuang parser to render qasm files using a python command.

AdriaanRol:
Won't fix. If this happens it will be part of OpenQL, not of PycQED.

PR #544, opened 2019-05-07
Assembler compatibility QCC and CCL

nyxus:
QCC and CCLight driver support for the separated assembler which is provided by the [UserEnv](https://github.com/DiCarloLab-Delft/UserEnv). Note that at this moment is not possible to include the QCC assembler (v4) and de CCL assembler (v2) in the same instance (shell). See https://github.com/DiCarloLab-Delft/UserEnv/issues/30 for more info of this problem. To support both the QCC and CCL this is (for now) the way to go.


pycqed/instrument_drivers/physical_instruments/QuTech_CCL.py
```diff
@@ -95,15 +95,7 @@ def _init_submodules(self):
         curdir = os.path.dirname(__file__)
         print(f"Assembler version {self.QISA.getVersion()}")
         if self.QISA.getVersion() == '4.0.0':
```
nyxus: Looking at this, it is probably better to do: `if self.QISA.getVersion() != '2. ...':`.

AdriaanRol:
@nyxus , could you make this PR against the QCC_testing branch? I want to ensure it works there as that is where we have everything working. 
I would also like to know if there is anything we need to change on the QCC setup to make this work.

nyxus:
@AdriaanRol, I've changed the base to `QCC_testing` branch. But there are many conflicts which I cannot resolve. The `Assembler_compatibility_QCC_CCL_final` was branched from `develop`. 

I'm not sure about our current QCC setup. I thought you did something special with the QCC assembler on that system (@MiguelSMoreira?). 
If https://github.com/DiCarloLab-Delft/UserEnv/pull/31 is accepted the setups should work with both assemblers (except from the issue described in https://github.com/DiCarloLab-Delft/UserEnv/issues/30).  But as far as I know, the current QCC test setup still has an early version of the UserEnv which requires the user to activate the environment from an admin powershell.  We should come together and discuss what to do (@AdriaanRol, @MiguelSMoreira and @nyxus) regarding the test setup.

Note 1, we also want to have the QCC stuff into the `develop` branch so we can start preparing the other measurements PCs. 
Note 2: we could not test on the QCC yet because it is used in experiments at this moment

MiguelSMoreira:
@nyxus and @AdriaanRol
There was nothing done to the assembler currently on the setup used to test QCC. The only modifications done to that computer were to OpenQL.

To ensure that we get a develop branch completely up to date, we should merge changes done to OpenQL to support QCC (guaranteeing that these won’t break support for CC-Light, Adriaan should be the best suited to determine this), and update both the QCC (and CCL) drivers and assemblers with the changes proposed in Assembler_Compatibility_QCC_CCL.

Given the number of different updates we would be making simultaneous, this should be done at a moment where disruption to experiments could be minimal and should include thorough testing of the resulting environment with both CCL and QCC.

PS: let’s not forget that, to the best of my understaing, the QCC testing branch is the only one that supports multiple feedline readout, since I believe it was on that computer that Niels worked on this

AdriaanRol:
@MiguelSMoreira , I understand @jacobdesterke ran into problems with this merge not being concluded yet yesterday. Can you take the time to make sure this PR is clean (as requested in previous comment)?

MiguelSMoreira:
Dear @AdriaanRol @jacobdesterke ,
As previously discussed with Adriaan, I have created a new branch with the modifications for CCL/QCC assembler compatibility, which I have successfully merged to develop under Pull Request #555 
Furthermore, I have removed this branch, as it is incorrect and no longer necessary.

I should finish by saying that the changes performed on both CLL and QCC drivers reflect the changes to the assemblers recently merged by me in Pull Request #31 of the UserEnv repository. Therefore, these are now consistent.

PR #547, opened 2019-06-06
Mock qubit

Timo1104:
Work on automatisation and mock qubit


AdriaanRol:
- Tests are failing because of qutip dependency. 
- Includes a new resonator power analysis. 
        This checks if there is a resonator power shift. 
       Determines size of dispersive shift and predicts qubit frequency. 
        Determines readout power. 
        Has no test. 
- New Initial_VNA_analysis 
        - Finds peaks/dips in VNA traces 
- New Spectroscopy power analysis that determines optimal spec power. 

- New qubit object calibration methods 
         - calibrate resonator power 
         - find qubit sweet spot 
- Find anharmonicity estimate (we may want to remove this).

PR #572, opened 2019-08-06
Move to stable qcodes from PyPI in CI

slavoutich:
QCoDeS version in CI was ancient, this fixes it.


slavoutich:
Polished a bit, should be slightly better now.

Issue #140, opened 2017-03-10
Standardized J2 chevron analysis


AdriaanRol:
@elrama- has made a really cool analysis that does an FFT on a chevron to extract J2. This is currently a quite big notebook. This should be included as a standard (simple ) measurment analysis script. 

It currently provides the following two figures 
![image](https://cloud.githubusercontent.com/assets/6142932/23787497/b5efccd2-0572-11e7-9d46-91640cf0c965.png)

And proceeds by doing a fit on the extracted points to extract J2.

Issue #132, opened 2017-03-03
More meaningful linecuts in 2D live plotting 


AdriaanRol:
Currently the plotmon for 2D plotting shows linecuts all in the same color, making it hard to distinguish the different lines. 
It would be useful to make this a bit more fancy to enhance the contrast. 
I'll write some ideas down here. 
- Different color for each linecut 
- some interactive tool to select which one one wants to display 
- Only show last-N linecuts and fade them out (sort of like a persitence mode).

MiguelSMoreira:
Closed due to inactivity

Issue #128, opened 2017-03-02
Standardized decoupling matrix calibration, analysis and flux control 


AdriaanRol:
This consists of puting the existing routines that are now in notebooks and done manual in easy to find places. It also involves updating the Flux control. 

- [ ] Experiment to measure the decoupling matrix 
- [ ] Analysis scripts for extracting the matrix 
- [x] Update the Flux control so that it obeys these conventions

MiguelSMoreira:
Closed due to inactivity

Issue #126, opened 2017-03-01
Migration of setups to beloved terminal


elrama-:
We have agreed amongst several people that we should move back to running experiments from the terminal. The main issue right now is that the PyQTgraph versions installed on the computers don't allow for this.

After installation of the whole software in the 2 new measurement computers(In the Aprilia section) we checked that this is working now, and ready to migrate to the terminal.

The procedure is as follows:
- Update anaconda to 3.6
- install the new pyqtgraph: Currently the conda install command has an old version. Use the PIP INSTALL command, and it is tested to work. In case you are reading this a loooong time from today, the versions used in the test are attached at the end.

Setups remaining to migrate:
 - [ ] Maserati
 - [x] Maserati Probe

Package's version:
alabaster (0.7.9)
anaconda-client (1.6.0)
anaconda-navigator (1.4.3)
astroid (1.4.9)
astropy (1.3)
Babel (2.3.4)
backports.shutil-get-terminal-size (1.0.0)
beautifulsoup4 (4.5.3)
bitarray (0.8.1)
blaze (0.10.1)
bokeh (0.12.4)
boto (2.45.0)
Bottleneck (1.2.0)
cffi (1.9.1)
chardet (2.3.0)
chest (0.2.3)
click (6.7)
cloudpickle (0.2.2)
clyent (1.2.2)
colorama (0.3.7)
comtypes (1.1.2)
conda (4.3.13)
configobj (5.0.6)
contextlib2 (0.5.4)
coverage (4.2)
cryptography (1.7.1)
cycler (0.10.0)
Cython (0.25.2)
cytoolz (0.8.2)
dask (0.13.0)
datashape (0.5.4)
decorator (4.0.11)
dill (0.2.5)
docutils (0.13.1)
et-xmlfile (1.0.1)
fastcache (1.0.2)
Flask (0.12)
Flask-Cors (3.0.2)
gevent (1.2.1)
greenlet (0.4.11)
h5py (2.6.0)
HeapDict (1.0.0)
idna (2.2)
imagesize (0.7.1)
ipykernel (4.5.2)
ipython (5.1.0)
ipython-genutils (0.1.0)
ipywidgets (5.2.2)
isort (4.2.5)
itsdangerous (0.24)
jdcal (1.3)
jedi (0.9.0)
Jinja2 (2.9.4)
jsonschema (2.5.1)
jupyter (1.0.0)
jupyter-client (4.4.0)
[...]

PR #576, opened 2019-08-09
Feature/z iupdate drivers

### LINKED ISSUE #92: UHFQC driver overwrites instrument get function
The UHFQC driver overwrites the Instrument get function used to get parameter values. 

The overwritten function is there to get things from the ZI API. Renaming this function should resolve the conflict.

nielshaandbaek:

Fixes #92, #128, #123, #137, #140, #126, #132, #111, #96.

- ZI drivers restructured such that the functionality to support dynamic waveform upload is managed by the ZI_base_instrument class. The driver monitors the waveforms that are configured using the dedicated waveform parameters. If it detects a change in length of a waveform, then it will trigger a recompilation of the AWG program next time the start() method is called. If a waveform is changed, but the configuration remains the same otherwise, then the waveform will be written directly to the instrument prior to the next time the AWG is started. The base class has also been extended with functionality to check for sufficient firmware/bitstream versions.
- A mock version of the server has been introduced, which is used in two very basic tests: one for the AWG-8 and one for the UHFQC driver. The test just instantiates the corresponding driver in the hope of triggering exceptions in case something is not working.
- UHFQC driver updated to use a more standard format for the AWG programs. It now uses a number of object variables to keep track of the kind of AWG program currently configured on the instrument. If the user tries to modify parameters that are not supported by the current program, then an error will be raised. In addition, it inherits the new waveform management functionality from the base class. Waveforms are no longer encoded in the AWG program, but written to CSV files prior to compilation.
- AWG-8 driver updated and simplified to use the new dynamic waveform uploading functionality provided by the base class.
- The LutMan's have also been updated (read-out, microwave and flux) to work with the new drivers. The LutMan's no longer have the dedicated real-time uploading methods. Instead, the user should make use of the standard `load_waveform(s)_onto_AWG_lookuptable` for changing the waveforms. Note that the waveforms will only be updated on the instrument the next time the AWG is started. The available tests for the LutMan's have been updated and should pass again.

It would be great if you could have a look at the changes @AdriaanRol.


nielshaandbaek:
Too many tests failed. Closing.

PR #577, opened 2019-08-12
Enh/data extraction

AdriaanRol:
new data extraction function that is simpler and more direct to use. 

For now I will use this when extracting coherence data. It is useful to also start integrating this in mav2 analyses but that needs some more thoughs. 

- @tstavenga , @elrama- The function to check is `pycqed.measurement.hd5_data.extract_pars_from_datafile`
    - There is an example in the tests on how to use it. 
    - Please let me know if the docstring is clear enough to understand how this works. 

- Another function has been added to the a_tools to find a datafile based on a timestamp. 
     -`pycqed.analysis.analysis_toolbox.get_datafilepath_from_timestamp`


elrama-:
The tests are failing at the function  data_from_time(timestamp, folder=None)


pycqed/analysis/analysis_toolbox.py:910: in get_folder folder = data_from_time(timestamp)

elrama-:
good. Everything passed, so I will merge

PR #585, opened 2019-09-16
Victor dev

caenrigen:
- refactored code to match CZ simulation notebook, only works for NE
- implemented circular interpolation and color scale for 'deg' units

@AdriaanRol


pycqed/measurement/waveform_control_CC/waveforms_flux.py
```diff
@@ -13,6 +13,7 @@
 import scipy.interpolate
 import numpy as np
 
+logger = logging.getLogger(__name__)
```
AdriaanRol: This should be `log = logging.getLogger(__name__)`

Issue #155, opened 2017-03-27
Add qumis files to hdf5 


AdriaanRol:
When using the CBox (or in the future the CC), it would be useful to save the qumis instructions that were uploaded in the datafile. 

Easiest way to do this is to make the contents of the file a parameter of the CBox. Need to figure out if that is the nicest way with datasaving (and monitoring the device etc).

wvlothuizen:
Cbox is dead

Issue #152, opened 2017-03-23
Split of user code (scripts)


AdriaanRol:
This relates to a discussion that has been going on for a while now. Recently discussed this with @elrama- and @brianzi . It should allow us to increase our code quality by having pull requests not cluttered with most of the daily script changes. It should also make everything a lot more readable (as in less of #144 ).

AdriaanRol:
I have just created a repository at https://github.com/DiCarloLab-Delft/PycQED_scripts that is accessible for group members. I'll send a proper update with instructions after I have moved over the files.

PR #598, opened 2019-11-22
Updated to not disconnect the instruments when 'close' is called. Now…

nielshaandbaek:
… the 'close' method will still work even if the server loses connection to the instrument. Also updated to not stop the running AWGs when 'close' is called (issue 155). Finally, a test was added for the new close behavior. The error stack is now cleared after the DIO calibration completes (in case it completes successfully).

Fixes QuSurf issue #155, #163 and #152.

Updated the 'close' method to not disconnect the instrument from the server. This has the added benefit of solving the problem with the 'close' method not working if the server loses connection to the instrument. Also updated the DIO calibration procedures to clear the error stack after calibration completes successfully so that the user is not swamped by the DIO timing errors detected during calibration. Added a small test of the new close behavior.

@AdriaanRol, @MiguelSMoreira

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_HDAWG8.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_HDAWG8.py
@@ -115,6 +115,9 @@ class ZI_HDAWG8(zicore.ZI_HDAWG_core):
         super().__init__(name=name, device=device, interface=interface, server=server, port=port, num_codewords=num_codewords, **kw)
         # Set default waveform length to 20 ns at 2.4 GSa/s
         self._default_waveform_length = 48
+        
+        # Holds the DIO calibration delay
+        self._dio_calibration_delay = 0
 
          # show some info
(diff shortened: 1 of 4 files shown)


MiguelSMoreira:
Everything seems ok to merge

pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_HDAWG8.py
```diff
@@ -749,5 +749,8 @@ def calibrate_CC_dio_protocol(self, CC, verbose=False, repetitions=1):
 
         # And configure the delays
         self.setd('raw/dios/0/delays/*', min_valid_delay)
-        # If succesful return True
+
+        # If succesful clear all errors and return True
+        self.clear_errors()
```
MiguelSMoreira: It would be better to ensure that only clear errors related to DIO calibration are cleared after calibration, as this may lead a user to overlook any other potential problems after calibration. 
Nonetheless, I believe this is not a blocking problem for now.

PR #600, opened 2019-11-26
Cleanup of Zi drivers

wvlothuizen:
- removed duplicates between ZI_HDAWG8.__init__ and ZI_HDAWG_core.__init__
- improved grouping of related functionality 
- replaced printf by logging
- introduced f-string literals for improved readability (and speed)

The changes in UHFQuantumController.py are difficult to track, because there are a lot of moves. The best I could find was to paste both versions into https://www.diff-text.com/diff-text-combined-navigation.aspx, and tick option 'don't show identical paragraphs'. Note that branch 'develop' contained function snapshot_base twice, so one is removed.

Can you have a look @MiguelSMoreira


pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
```diff
@@ -383,25 +333,6 @@ def _check_versions(self):
     def _num_channels(self):
         return 2
 
-    def _reset_awg_program_features(self):
```
MiguelSMoreira: It appears that the _reset_awg_program_features method was removed after the code refactoring. I do not know if this was intended or not, or even whether we are currently making use of the functionality implemented by this method

PR #612, opened 2020-01-27
add residual ZZ analysis

baklaver:
- Add residual crosstalk analysis to analysis_v2
- This analysis finds the frequencies of multiple residual ZZ measurements by fitting a damped oscillator function. 
- The found frequencies are plotted in a single colored matrix as shown in the example below. 
residual_matrix_20200124_182258


AdriaanRol:
@baklaver could you add a test that serves as an example on how to use this analysis?

pycqed/analysis_v2/residual_ZZ_analysis.py
```diff
@@ -0,0 +1,192 @@
...
+                                  value=1,
+                                  vary=False)
+    params = damped_osc_mod.make_params()
+
+    fit_res = damped_osc_mod.fit(data=y,
+                                 t=x,
+                                 params=params)
+    return fit_res
+
+def plot_crosstalk_matrix(crosstalk_matrix,
```
AdriaanRol: This function could use a docstring. 
What is the crosstalk matrix? 
What are the input arguments? 
Are there assumptions on the shape of the crosstalk matrix (e.g., has to be n*n) and the units? 
Same for qubit_labels, what are "combinations"? 
Is there a way to specify the limits to the cbar?

pycqed/analysis_v2/residual_ZZ_analysis.py
```diff
@@ -0,0 +1,192 @@
...
+                    va='center', ha='center', color=col)
+
+    ax.set_xticklabels(valid_combinations)
+    ax.set_xticks(np.arange(len(valid_combinations)))
+    ax.xaxis.set_ticks_position('bottom')
+
+    ax.set_yticklabels(combinations)
+    ax.set_yticks(np.arange(len(combinations)))
+    ax.set_ylim(len(combinations)-.5, -.5)
+    ax.set_ylabel('Echo qubit')
```
AdriaanRol: I don't understand what these labels mean. 
Echo qubit and control qubit have some specific meaning in the experiment but that is not clear here. It may still be a good label but it definitely needs a comment to explain it.

pycqed/analysis_v2/residual_ZZ_analysis.py
```diff
@@ -0,0 +1,192 @@
...
+import lmfit
+from pycqed.analysis import fitting_models as fit_mods
+from pycqed.analysis.tools.plotting import set_xlabel, set_ylabel, \
+    cmap_to_alpha, cmap_first_to_alpha
+
+class Residual_Crosstalk(ba.BaseDataAnalysis):
+    """
+    Residual crosstalk analysis.
+
+    Produces residual crosstalk matrix from multiple crosstalk measurements.
```
AdriaanRol: What are crosstalk measurements? 
Crosstalk is a widely used but ill defined concept. One should explain what experiment it uses either through a reference or by explaining it here.

pycqed/analysis_v2/residual_ZZ_analysis.py
```diff
@@ -0,0 +1,192 @@
...
+                'matrix': residual_ZZ_matrix,
+                'qubit_labels': self.qubits}
+    def prepare_plots(self):
+        self.plot_dicts['residual_matrix'] = {
+            'plotfn': plot_crosstalk_matrix,
+            'crosstalk_matrix': self.proc_data_dict['quantities_of_interest']['matrix'],
+            'qubit_labels': self.qubits
+        }
+
+def fit_residual_coupling_oscillation(x, y, **kw):
```
AdriaanRol: So you are fitting residual coupling oscillations. 
Why are we not plotting this? Or creating a range of plots. 
I think those underlying plots are extremely important. 
Whenever we look at the matrix you produce and see something weird, the first thing to do is a sanity check, does this make sense? Right now that is virtually impossible.

AdriaanRol: Related to this, one can now not run this residual ZZ analysis on a single coupling, only on a grid containing multiple ones.

pycqed/tests/analysis_v2/test_residual_ZZ_analysis.py
```diff
@@ -0,0 +1,39 @@
...
+		ts_list = ma.a_tools.get_timestamps_in_range('20200124_182258','20200124_183814', '')
+		a = rza.Residual_Crosstalk(qubits=qubit_labels, t_start=ts_list[0], t_stop=ts_list[-1])
+		for i in range(len(qubit_labels)):
+			self.assertTrue(np.isnan(a.proc_data_dict['quantities_of_interest']['matrix'][i,i]))
+
+	def test_matrix_non_diagonal(self):
+		# Test if non-diagonal matrix elements are close to expected data values
+		qubit_labels = ['D1', 'Z1', 'X', 'D3']
+		ts_list = ma.a_tools.get_timestamps_in_range('20200124_182258','20200124_183814', '')
+		a = rza.Residual_Crosstalk(qubits=qubit_labels, t_start=ts_list[0], t_stop=ts_list[-1])
```
AdriaanRol: OK, so here I see that one needs a range of timestamps. 
What are the conditions these timestamps need to have in order to be able to run the analysis? These conditions for when the analysis works are not documented anywhere.

AdriaanRol:
@baklaver @elrama- 
FYI, worked out the (non existing) factor of 2. 
Note that this note also defines the sign of $\zeta$. 
image

MiguelSMoreira:
Status update @baklaver @elrama- ?

MiguelSMoreira:
Do you have an update on this pull request @elrama- ?

MiguelSMoreira:
Closed due to lack of activity!

Issue #146, opened 2017-03-13
Confusion about integration_logging_det and integrated_average_detector


antsr:
As far as I understand from the code the integration logging detectors are essentially the same as the integrated average detectors, only that they do not do any averaging and they take the number of segments as an initialization parameter instead of from the length of the sweep function, which they ignore. So could an integrated average detector be always used instead of integration logging detector, by setting the number of averages to 1 and passing as `sweep_points` a vector of length equal to the number of shots desired? In my opinion it would improve code readability even more, if in such case the sweep function was set to a `Hard_Repeat` sweep function which wouldn't have to do anything, and the AWG sequence was loaded by calling the sequence function directly.

AdriaanRol:
@antsr , your observation is correct. The logging detector is an averaging detector without averaging. However, considering the diverse use cases (as in the type of experiments these are used for) and the different way they interact with the acquisition loop I think this is different enough to warrant a separate detector. 

Feel free te reopen the issue if you have any further question or disagree.

PR #631, opened 2020-06-19
Log file creation for debugging purposes and alternative data acquisition mode

nielshaandbaek:
This pull request adds two features to the ZI drivers. 

First, a new argument is introduced when creating the driver called `logfile`. If this argument is set (default is not to set it) then a text logfile will be generated where all commands sent to the device are logged. The file format is such that it can be executed in a Python notebook where the original objects are available. The goal is to assist with debugging.

Second a new acquisition mode is added which aims to resolve Git issue #146. The `acquisition` functions now take a new argument `poll` with a default value of `True`. With the default value the functionality is unchanged. If the parameter is set to `False` then the function will instead of polling for data wait for the AWG to finish and then manually read out the data. Note that the functions called by `acquisition` can also be used individually as it is done in the detector functions. By not subscribing to data and instead simply waiting until the AWG is done we avoid the warnings with the data server dropping data values.

Fixes https://github.com/DiCarloLab-Delft/QuSurf-IssueTracker/issues/146.

@antsr 

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
@@ -759,16 +759,19 @@ setUserReg(4, err_cnt);"""
     ##########################################################################
 
     def acquisition(self, samples=100, averages=1, acquisition_time=0.010, timeout=10,
-                    channels=(0, 1), mode='rl') -> None:  # FIXME: wrong return type
+                    channels=(0, 1), mode='rl', poll=True):
         self.timeout(timeout)
-        self.acquisition_initialize(samples, averages, channels, mode)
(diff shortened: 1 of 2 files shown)


pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
```diff
@@ -878,12 +886,46 @@ def acquisition_poll(self, samples, arm=True,
...
+    def acquisition_get(self, samples, arm=True,
+                         acquisition_time=0.010):
+        """
+        Waits for the UHFQC to finish a measurement then reads the data.
+
+        Args:
+            samples (int): the expected number of samples
+            arm    (bool): if true arms the acquisition, disable when you
+                           need synchronous acquisition with some external dev
+            acquisition_time (float): time in sec between polls
```
antsr, line 898: This unused argument should probably be removed from the function

PR #641, opened 2021-02-08
Added PQSC driver from ETH

### LINKED ISSUE #180: Detectors for UHFQC correlations mode

nielshaandbaek:
This PR adds the PQSC driver from ETH and improves it with proper error handling. It also adds the standard `start` and `stop` methods, which would be the preferred way of starting the PQSC. Secondly, the PR updates the `ZI_base_instrument` class to support instruments such as the PQSC that does not have an AWG.

Fixes #180.

- Added ZI PQSC driver
- Updated ZI_base_instrument to work with instruments without an AWG.

@chellings  the name of someone you want to review this pull request. 

- travis test suite passes

- follow the PEP8 style guide 
- have tests for the code

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_PQSC.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_PQSC.py
@@ -11,6 +11,7 @@ import logging
 import numpy as np
 import pycqed
 import json
+import copy
 
 import pycqed.instrument_drivers.physical_instruments.ZurichInstruments.ZI_base_instrument as zibase
 
@@ -161,7 +162,7 @@ class ZI_PQSC(zibase.ZI_base_instrument):
         found_errors = False
 
         # Combine errors_to_ignore with commandline
-        _errors_to_ignore = self._errors_to_ignore
+        _errors_to_ignore = copy.copy(self._errors_to_ignore)
         if errors_to_ignore is not None:
             _errors_to_ignore += errors_to_ignore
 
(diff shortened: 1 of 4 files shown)


MiguelSMoreira (approved):
Everything looks good! 
Do I have your permission to merge? Or does this branch pertain to ongoing development?

pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_PQSC.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            raise_exceptions = True
+
+        # Asserted in case errors were found
+        found_errors = False
+
+        # Combine errors_to_ignore with commandline
+        _errors_to_ignore = self._errors_to_ignore
+        if errors_to_ignore is not None:
+            _errors_to_ignore += errors_to_ignore
```
chellings, line 166: Note that _errors_to_ignore is a reference to the mutable self._errors_to_ignore. This means that when calling
```
PQSC.check_errors(errors_to_ignore=['SOMEERROR'])
```
once, SOMEERROR will always be ignored afterwards. I guess that this is not the intended behavior and that something like
```
_errors_to_ignore = self._errors_to_ignore + errors_to_ignore if errors_to_ignore is not None else self._errors_to_ignore
```
should be used, right? (Or, if this is the intended behavior, it should be clarified in the docstring because it is quite unexpected from the user's perspective.)

The same issue exists in ZI_HDAWG_core.check_errors

chellings (changes requested):
Thank you Niels for adding and improving the PQSC driver.

Please check whether you agree with my additional commit, and please check my question about _errors_to_ignore.

Thanks a lot!

chellings:
Hey @nielshaandbaek I have looked at the new commits that you have force-pushed, but it seems that they are the same commits as before, i.e., the bug with _errors_to_ignore is still there.

nielshaandbaek:
I finally updated the HDAWG too.

chellings (approved):

PR #649, opened 2021-06-23
add pre-commit configuration

eendebakpt:
Changes proposed in this pull request:
- Add pre-commit configuration file

@wvlothuizen The [pre-commit](https://pre-commit.com/) configuration add automatic formatting to all committed files using `autopep8` and `pyupgrade`. More tools can be added, see for example the config file for [qtt](https://github.com/QuTech-Delft/qtt/blob/dev/.pre-commit-config.yaml). For pycqed3 this initial configuration is quite conservative.

PR #652, opened 2021-06-25
[RFC] autopep8

eendebakpt:
- Apply autopep to python files in the repo

Using automatic formatters keeps code clean and allows for a uniform coding style. This PR applied autopep8 (basic settings, but with max line length 120). This is in line with the issue template and contributing guidelines which prefer autopep8 (https://github.com/DiCarloLab-Delft/PycQED_py3/blob/develop/.github/CONTRIBUTING.md)

Related is #649 which adds configuration files to automatically format according to autopep8 on commiting.

Once we agree, we will autopep all files in the repository.

@wvlothuizen @MiguelSMoreira

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/MinicircuitsRC.py
+++ b/pycqed/instrument_drivers/physical_instruments/MinicircuitsRC.py
@@ -10,45 +10,48 @@ from qcodes.utils import validators as vals
 
 from pyvisa import VisaIOError
 
+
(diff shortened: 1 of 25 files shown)


wvlothuizen:
Why limit this to physical_instruments?

eendebakpt:
@wvlothuizen Full autopep takes quite some time. This is a start to keep feedback on the format. If we agree, then I will do a full autopep8.

PR #660, opened 2021-10-08
FIX CI failing on OpenQL 0.10.0 compatibility

wvlothuizen:


wvlothuizen:
just merged, this is so small

PR #666, opened 2021-11-17
FIX CI (once more)

wvlothuizen:


MiguelSMoreira (approved):
Indeed, it seems better to skip these tests than to leave CI failing. Thank you for the other fixes as well!

PR #620, opened 2020-02-19
[WIP] PR to monitor whether cleanup breaks anything

wvlothuizen:
The purpose of this PR is to get rid of unused stuff, in a reversible way. Unused code makes it more difficult to maintain the code base, and reduces code coverage since for many old pieces of code no tests exist.

PR #669, opened 2021-11-18
further cleanup

wvlothuizen:
Prior deprecations (PR #620 and #646) moved unused files to a hierarchy under "/deprecated", and commented out parts of files that referred to these unused files. This scheme is used to ease reverting any deprecation if necessary.

With the release of PycQED version 0.3 as a solid base to fall back to, this PR now removes these commented out parts as a further stage of cleaning up


MiguelSMoreira (approved):
Looks good! Thanks for cleaning all of this up

PR #671, opened 2021-11-19
Group related methods

wvlothuizen:
This PR reorders methods in some core parts of PycQED to allow better insight in the inheritance structure (and quirks thereof). 
Because it is sort of impossible to track history when shuffling around code within files, this PR does not change any functionality. Some comments were added however in places where questionable things happen.

Note that other changes to the files involved should ideally not be carried over the merge of this PR (so merge any other changes in progress before this PR, and after this PR, pull develop before editing).

Subsequent steps are foreseen to improve LutMan and codeword handling.


MiguelSMoreira (approved):
As you said, it is quite hard to keep track of these sorts of changes but, to the best of my ability, all the code moving appears consistent and correct. This pull request is approved!

PR #675, opened 2021-12-13
split-off HW handling 

wvlothuizen:
This PR is intended to split off the hardware handling from the files below, as an initial step to handle hardware in a separate 'layer':
- CCLight_Transmon
- device_object_CCL
- detector_functions

The following still need a similar split-up:
- mw_lutman
- ro_lutman
- flux_lutman (which of the 3?)
- sweep_functions

Some class and file names have been changed to better reflect the current situation (e.g. "CCLight" is removed), but to maintain compatibility the old names and locations are still supported through 'compatibility imports'.

This PR takes a lot of care to not break anything for the end user, but in the slipstream of this PR:
- support for CCL/QCC was removed, since OpenQL no longer supports this architecture
- a lot of old cruft was cleaned up
- some long parameter lists are reformatted for improved readability
- some type annotations are added in order allow better understanding, improved checks and improved code navigation by PyCharm
- deprecation warnings were added to apparently unused code
- some disabled CI tests were fixed and enabled
- some CI deprecation warnings were solved 
- fixed codacy and codecov CI integrations

A subsequent step in a follow-up PR would be to integrate the HW management for single vs. multi qubit experiments, and more clearly assign the responsibilities of all the components involved


wvlothuizen:
@MiguelSMoreira I guess it would be advantageous to merge this before new developments will start to prevent new conflicts now things have moved around. 
I merged with the recent updates in branch 'develop' and fixed CI. It would be good to take a special look at the last few commits that fix conflicts and CI.
Maybe other people should also review, I'll leave that for you to decide.

pycqed/instrument_drivers/meta_instrument/device_object_CCL.py
```diff
@@ -1,7688 +1,7 @@
...
-            True if calibration succeeded, False if it failed.
-
-        procedure works by performing a conditional oscillation experiment at
-        a phase of 90 degrees. If the phase correction is correct, the "off" and
-        "on" curves (control qubit in 0 and 1) should interesect. The
-        analysis looks for the intersect.
-        """
-
-        if prepare_for_timedomain:
-            self.prepare_for_tim
```
MiguelSMoreira: I don't understand, if the entire contents of the device object were trivially migrated to the new hardware-abstraction layer (HAL), including all the measurement routines, what is then the point of the HAL? Shouldn't it encompass only abstraction of the interactions with hardware? 

If it encompasses also knowledge of the quantum device and calibration experiments, like is encoded into the qubit object and device object, isn't this just a change in branding? In which case this will further confuse people in the laboratory, who are used to the current structure of code when looking for and building new experiments.

pycqed/instrument_drivers/meta_instrument/qubit_objects/CCL_Transmon.py
```diff
@@ -1,7451 +1,6 @@
...
-                    self.T2_echo(a.fit_res.params['tau'].value)
-            elif sweep == 'order':
-                a = ma2.Single_Qubit_TimeDomainAnalysis(label='CPMG', auto=True, close_fig=True)
-
-            return a
-
-
-    def measure_spin_locking_simple(self, times=None, MC=None,
-                     analyze=True, close_fig=True, update=True,
-                     label: str = '', prepare_for_timedom
```
MiguelSMoreira: Again, I feel like including the full qubit objects (and the device object like mentioned above) in the hardware-abstraction layer is just needless re-branding of these files, which will simply contribute to confusion when working with them on the laboratory. What advantages does this bring?

MiguelSMoreira (commented):
In addition to the comments I left inlined, what is the advantage of rebranding (and moving) all the detector functions to this new detector functions folder? 

This might otherwise break various import statements across user scripts and functions...

MiguelSMoreira (approved):
With the previously mentioned caveats, this pull request is approved. Please feel free to merge (and delete the merged branch, if appropriate) when you feel appropriate @wvlothuizen

PR #677, opened 2022-01-03
Draft: add Zurich Instruments SHFQA driver

tonyp-zi:
Work in progress. I have just set up SHFQA files as copies of their UHFQA counterparts, and done some minor adjustements.

Todo:
- [ ] Adapt SHFQA_core and SHFQA class-level variables
- [ ] Adapt SHFQA_core._default_waveform_length
- [ ] Adapt functions in SHFQA.py and SHFQA_core.py as needed
- [ ] Add SHFQA to MockDAQServer?
- [ ] Adapt ZI_base_instrument.py as needed

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/SHFQA.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/SHFQA.py
@@ -30,6 +30,7 @@ log = logging.getLogger(__name__)
 ##########################################################################
 
 class SHFQA(shf.SHFQA_core, DIO.CalInterface):
+    # TODO(TP): Adapt to SHFQA
     """
     This is the PycQED driver for the 1.8 Gsample/s SHFQA developed
     by Zurich Instruments.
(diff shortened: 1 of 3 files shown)


wvlothuizen:
I'd like to make a strong case to not create a SHF driver by a copy/paste/edit of the UHF driver, because it will be a nightmare to maintain and integrate into the rest of PycQED. IMHO it would be highly desirable to identify the commonalities between the instruments, and handle those in common code.

I'm actually working on splitting of hardware dependencies throughout PycQED (PR #675), and interaction with the UHF happens in many places (hopefully to be drastically reduced in the future). Making the interface of the UHF and SHF as common as possible/sensible, and making the commonalities explicit, will be some extra work now, but will probably be paid back many times during integration, debugging and maintenance.

@tonyp-zi this PR was deleted while I was typing the comment above, so the comment may no longer be valid

tonyp-zi:
Hi @wvlothuizen, thank you for your quick feedback. That was just a very rough PR, to do some preliminary testing.

Your comment is very sensible, we will make sure to structure the driver properly to facilitate its maintenance. We still have to figure out exactly what can be shared between the two devices, and how to reflect it in the code, I will keep you updated.

@martinr-zi @mariosk-zi

PR #680, opened 2022-01-18
Further CC updates

wvlothuizen:
This PR provides:
- some updates to the OpenQL configuration for the CC
- speed improvements for the generation of RB sequences:
    - no longer reread hash tables from file on every call to two_qubit_clifford_group.py::get_clifford_id()
    - cache gate decompositions
    - compute pauli transfer matrices on demand, and cache them
    - example performance improvement (test changed to compute 2000 cliffords instead of normal 20):
       - original: `199.26s call     pycqed/tests/test_cliffords.py::TestCliffordClassRBSeqs::test_interleaved_randomized_benchmarking_sequence_2Q`
       - updated: `3.58s call     pycqed/tests/test_cliffords.py::TestCliffordClassRBSeqs::test_interleaved_randomized_benchmarking_sequence_2Q
`


requirements.txt
```diff
@@ -18,7 +18,7 @@ pyqtgraph
 matplotlib
 autodepgraph
 networkx
-qutechopenql
+qutechopenql>=0.10.3
```
MiguelSMoreira, line 21: Wouldn't this requirement make it harder for those that still use OpenQL for CCL (requiring <0.10.0)?  Granted pip would always install latest version of requirements regardless, but after manually downgrading qutechopenql, this would make the requirements for PycQED fail upon re-installation

MiguelSMoreira (approved):
I didn't necessarily see any breaking changes. This pull request has my go-ahead for merging!

requirements.txt
wvlothuizen: true, although you would need to build OpenQL manually anyway. The reason to include this is to not be satisfied with 0.10.0, the latest available on PyPi (0.10.3 is only available on https://github.com/DiCarloLab-Delft/OpenQL/releases)

PR #651, opened 2021-06-25
Added function to return minimum readout hold-off including unit test.

nielshaandbaek:
Fixes #191.

Changes proposed in this pull request:
- Added `minimum_holdoff` attribute to `UHFQuantumController` class.
- Added unit test of `minimum_holdoff` attribute.
- Attribute returns the minimum hold-off in units of seconds between two readout operations.

@chellings 

In order for the pull request to be merged, the following conditions must be met:
- test suite (Github actions) passes
- all reasonable issues raised by codacy must be resolved 
- a positive review is required

Whenever possible the pull request should
- follow the PEP8 style guide 
- have tests for the code
- be well documented and contain comments

Tests are not mandatory as this is generally hard to make for instruments that interact with hardware.

PR #681, opened 2022-01-28
Move minimum_holdoff from UHFQC to UHFQA_core (including unit test)

chellings:
This feature from #651 belongs into the UHFQA_core driver (lower-layer driver shared by TUD and ETH) that was introduced in #656. The feature ended up in the wrong place because the two pull requests were processed in parallel. This pull request moves the minimum_holdoff calculation to the correct class, and also adds the corresponding unit test originally written as part of #651  to the test module of the core driver.

@MiguelSMoreira

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQA_core.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQA_core.py
@@ -200,6 +200,14 @@ class UHFQA_core(zibase.ZI_base_instrument):
             'actual readout.',
             vals=validators.Ints())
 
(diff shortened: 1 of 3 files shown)


MiguelSMoreira (approved):
That was a good catch @chellings
Everything looks good and tests are passing, I will merge the pull request!

MiguelSMoreira:
I have deleted the branch, after having merged it into develop, to cleanup things. 
If it is still needed, please feel free to restore it!

chellings:
Thanks for reviewing & merging so rapidly @MiguelSMoreira. Yes, deleting the branch was fine.

PR #684, opened 2022-02-18
[WIP] local changes in ferrari setup

wvlothuizen:
just a test to see what we would get


wvlothuizen:
@MiguelSMoreira doesn't look too difficult to backport

MiguelSMoreira:
@wvlothuizen it was silly of me to attempt merging of develop onto local, with all the recent changes done on develop, should have been the other way around.

That's a good point, most of the changes made were for automated calibrations (some changes of thresholds, conditions, etc for common experiments) and preventing the use of matplotlib interactively. Most other changes are relatively minor. However, I am intrigued how GitHub is able to track the changes, for example, to CCL Transmon, here since that class was 'renamed' to HAL_Transmon. Is git smart enough to track changes across files that were renamed, or is this PR done against a not-up-to-date version of develop (doesn't appear to be the case)?

wvlothuizen:
Regarding the changes presented, note that Github normally shows a three-dot diff that can be confusing if you're not aware: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons

MiguelSMoreira (approved):
Everything looks good! These are all necessary changes related to plotting in interactive mode, recording calibration parameters and graph-based tune-up.

PR #689, opened 2022-03-21
Make dio calibration of UHFQC compatible with LabOne 22.02

### LINKED ISSUE #687: Make DIO calibration compatible with LabOne 22.02
In LabOne 22.02 the behavior of real-time (RT) logging changed with new public nodes introduced. The PycQED driver implementation of the calibration relied on raw nodes. Adapt the PycQED driver for LabOne 22.02

Acceptance criteria
- The DIO calibration can be performed the usual way also with LabOne 22.02.
- The adapted driver works for older LabOne versions in a way transparent to the user.
- Public nodes are used wherever possible.

martinr-zi:
Fixes fixes #687

@schenfab, @mariosk-zi @michelecollodo

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/UHFQuantumController.py
@@ -1039,10 +1039,20 @@ setTrigger(0);
         for delay in range(12):  # NB: 16 steps are available, but 2 periods of 20 ns should suffice
             log.debug(f'{self.devname}:    Testing delay {delay}')
             self.setd('raw/dios/0/delay', delay)  # in 1/300 MHz = 3.33 ns steps
+            try:
+                # LabOne 22.02 and higher: clear the sticky timing error detection bits
+                self.setd('raw/dios/0/error/timingclear', 0xffffffff)
+            except RuntimeError:
+                # no timingclear node available
+                pass
             time.sleep(0.5)
             valid_sequence = True
             for awg in [0]:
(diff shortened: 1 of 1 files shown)


mariosk-zi (approved):

schenfab (changes requested):
The nodes are not yet correctly used for the new scheme with >=22.02.

The old usage concept was along this line:
```
self.setd('raw/dios/0/delay', delay)
time.sleep(0.5)
[...]
error_timing = self.geti('raw/dios/0/error/timing')
```

The new concept should look like:
```
self.setd('raw/dios/0/delay', delay)
self.setd('raw/dios/0/error/timingclear', 0xffffffff) # clear the sticky timing error detection bits
time.sleep(0.5)
[...]
error_timing = self.geti('raw/dios/0/error/timingsticky') # readout timing errors that accumulated since the last call to "timingclear"
```

The important improvement here is the addition of the "timingclear" node, which is required to be manually set before checking for timing errors (by sleeping some time and reading the sticky timing error node afterwards). Previously, this clearing was done automatically by the device at arbitrary points in time, meaning reading the timing error node could have lead to false negatives (one would have got no timing errors, while actually there were some that were auto-cleared by the device just before reading the node).

Long story short: You first need to figure out if the old or new nodes are present, and then conditionally also set the "timingclear" node before sleeping... Probably by adding yet another try construct for the "timingclear" (except case would `pass` here), or maybe a bit more nicely, use listNodes beforehand to figure out which scheme is used and get rid of the try/excepts...

schenfab (approved):
LGTM, thanks for the change!

wvlothuizen (approved):
Looks good

PR #690, opened 2022-03-22
Martinr/shfqa driver

martinr-zi:
First draft of an SHFQA driver reproducing as best as possible the existing interface of UHFQA_core + UHFQC

- introduce dependency to zhinst-deviceutils package (https://pypi.org/project/zhinst-deviceutils/)
- add parameter files for SHFQA2 and SHFQA4
- compute awg index via dedicated method in ZI_base_instrument instead of hardcoded ch // 2
- SHFQA code

WARNING:
- The code is not yet sufficiently tested to be used in a production environment
- There is no support yet for DIO calibration

@tonyp-zi

### DIFF
--- a/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_base_instrument.py
+++ b/pycqed/instrument_drivers/physical_instruments/ZurichInstruments/ZI_base_instrument.py
@@ -916,7 +916,7 @@ class ZI_base_instrument(Instrument):
                 par_kw['set_cmd'] = _gen_set_cmd(self.setv, parpath)
                 par_kw['get_cmd'] = _gen_get_cmd(self.getv, parpath)
                 # min/max not implemented yet for ZI auto docstrings #352
-                par_kw['vals'] = validators.Arrays()
+                par_kw['vals'] = validators.Arrays(valid_types=(complex, np.integer, np.floating))
 
             elif par['Type'] == 'String':
                 par_kw['set_cmd'] = _gen_set_cmd(self.sets, parpath)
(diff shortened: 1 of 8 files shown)


tonyp-zi:
Good job!

tonyp-zi:
It is not clear to me, but we might be missing a complex conjugation somewhere. I think we want to make sure that the integration is (I_w - i Q_w) (I_s + i Q_s). Maybe we should check with Michele.

MiguelSMoreira (approved):
Everything looks good for some preliminary testing!

Quick question @martinr-zi, particularly regarding the shfqc_uhfqc compatibility layer, I see the wrapper function for setting resonator combinations from codewords. The setting of resonator frequencies will be done directly from a settable/gettable node? Is this what is referred to by digital oscillator frequencies in the node-doc files?

martinr-zi:
@MiguelSMoreira thanks for the feedback! 

As of now, the resonator frequencies must be set in two steps: the RF center frequency is set through a dedicated device node as follows:

`driver.set("qachannels_{ch}_centerfreq", value)`

The IF offset of the maximum 7 Qubits or slots attached to a certain channel is then set through the waveform samples uploaded to the device (readout pulse and integration weights). For the readout pulses, its the codeword waveforms, for the integration weights, this is done either through manually uploading a complex waveform through 

`driver.set(f"qachannels_{ch}_readout_integration_weights_{slot}_wave", weight)`

or by using the helper function ported from the UHFQC, `prepare_SSB_weight_and_rotation()`, which is more restrictive in what it allows you to have as integration weights.

Does that answer your question?

MiguelSMoreira:
@martinr-zi, yes, it does. Thank you for clarifying!

PR #700, opened 2022-05-02
Python 3.7+ support (including 3.8 and 3.9)

### LINKED ISSUE #692: Upgrade to Python 3.7
Python 3.6 has reached end-of-life (see https://www.python.org/downloads/release/python-366/), and only supports old versions of QCoDes.
So we should move on, note that CI already tests on 3.7 in addition to 3.6

wvlothuizen:


MiniSean:
- [X] Successful (manual) hardware integration test. Should be ready to be reviewed.

MiniSean (approved):
These changes were verified on one of the measurement setups (Aprilia 30-March-2023).
Using frequency and (single qubit) time domain experiments.

PR #703, opened 2022-06-15
fixes for issue #702

### LINKED ISSUE #702: CC driver functions checking DIO calibration are broken
During debugging by ZI, it appeared that the following functions in CCCore.py are broken:
- `get_calibrate_dio_success()`
- `get_calibrate_dio_read_index()`

They ignore the `ccio` parameter,  and when that's fixed, give erroneous results. The previous method of using `debug_get_ccio_reg()` does seem to work correctly (but the register indices may change with firmware versions)

wvlothuizen:


wvlothuizen:
merged small change without review

PR #705, opened 2022-06-23
Hide dependency of CC to shfqa driver using magic number mask

martinr-zi:
Use a magic number for the dio mask used in calibration of the SHFQA instead of introducing a programmatic dependency to the SHFQA driver. This fixes the breaking build for the central controller for 3.6.


wvlothuizen (approved):
perfect for now.
At some point we could compute the mask (and those for different DIO modes) from a mode specification in DIO.py (since magic numbers are indeed a bad thing as you pointed out)

PR #709, opened 2022-07-07
Martinr/adapt shfqa to single channel uhfqa compatibility mode

martinr-zi:
Adapt shfqa to single channel uhfqa compatibility mode

https://github.com/DiCarloLab-Delft/QuSurf-TestingFramework/projects/1#card-82962031


wvlothuizen (approved):
Looks good

PR #710, opened 2022-07-19
Add dio parity checks for HDAWG

martinr-zi:
This PR implements support for DIO parity checks on the ZI HDAWG at the PycQED driver level. 

Related issue: https://github.com/DiCarloLab-Delft/QuSurf-IssueTracker/issues/211

- Override base class methods checking for errors to also account for DIO parity check errors
- Ignore errors generated during DIO calibration


martinr-zi:
I propose to postpone the merge until we have a ready-to-go patch release that was successfully tested on the shared test rack.

PR #721, opened 2023-01-25
UHF multi detector: implement fast poll of multiple UHFs

ysalatheZI:

Loop over detectors in the inner-most poll loop in the UHF multi detector.

@leodicarlo , @mariosk-zi 

- test suite (Github actions) passes

- follow the PEP8 style guide 
- have tests for the code

### DIFF
--- a/pycqed/measurement/det_fncs/hard/UHFQC.py
+++ b/pycqed/measurement/det_fncs/hard/UHFQC.py
@@ -44,6 +44,22 @@ class Multi_Detector_UHF(Multi_Detector):
         for detector in self.detectors:
             new_values = detector.get_values(arm=False, is_single_detector=False)
             values_list.append(new_values)
+
+        # Pad all result vectors for them to have equal length.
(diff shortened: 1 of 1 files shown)


ysalatheZI:
This will require more work to properly prepare and post-process the data for the different detector classes.

ysalatheZI:
This pull request is ready for review.

MiniSean:
@RuggeroDiCarlo and @leodicarlo, what is the status of these changes?
Have they already been tested on the Inspire setup?

MiniSean:
After talking with @RuggeroDiCarlo I understand that there are two (identical) instances of this code.
One existing in this PR, and one already existing on the Inspire setup (not yet pushed to the `inspire_detached` branch). 
After doing a branch comparison (for the relevant file `UHFQC.py`) both are indeed identical and working as intended.

Naturally I rather not push changes this way since it is very sensitive to errors and merge conflicts.
But we can make an exception for this PR.

PR #727, opened 2023-10-23
Replacing deprecated functions

msamiotis:
Necessary fixes to imports, where deprecated functions/methods were being used. All those have been replaced with working ones. Necessary changes for the SHFPPC calibration script to run.


MiniSean (approved):
Replacement of `qcodes` by `qcodes_loop` is correct. This change also needs to be made to CI, which right now is not recognizing `qcodes_loop`.

PR #736, opened 2025-07-31
Merging July 2025 changes made by Ruggero and Sean

msamiotis:
Changes done are mostly for the timing calibrations, and the option to park some qubits during some of the 2Q gate calibration measurements.

### DIFF
--- a/pycqed/instrument_drivers/meta_instrument/HAL_Device.py
+++ b/pycqed/instrument_drivers/meta_instrument/HAL_Device.py
@@ -411,6 +411,10 @@ class HAL_Device(HAL_ShimMQ):
         MC.set_sweep_points(p.sweep_points)
         d = self.get_int_avg_det()
         MC.set_detector_function(d)
+        
+        # Check low frequency qubit CZ amplitude
+        print("Amplituce SE", self.find_instrument("flux_lm_SE").vcz_amp_dac_at_11_02_NE())
+        
         MC.run(
             "conditional_oscillation_{}_x{}_{}{}".format(
                 list_qubits_used, cz_repetitions,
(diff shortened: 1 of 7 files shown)


pycqed/instrument_drivers/meta_instrument/HAL_Device.py
```diff
@@ -6453,7 +6459,9 @@ def measure_vcz_A_tmid_landscape(
         flux_codeword: str = 'cz',
         flux_pulse_duration: float = 60e-9,
         prepare_for_timedomain: bool = True,
-        disable_metadata: bool = False):
+        disable_metadata: bool = False,
+        TLS: bool = False # added by RDC 21-07-2025
```
MiniSean: I talked to Ruggero that I recommend renaming this argument (from `TLS`) into` `force_lsq_no_detuning`.
The default behaviour of this function was to set the pulse amplitude of the lower-frequency qubit (lsq) to 0.
However, when performing two-qubit gates, sometimes we want to move the entire interaction point. For this we have to set the pulse amplitude of the lower-frequency qubit (lsq) to a non-zero value based on `vcz_amp_dac_at_11_02`.

pycqed/instrument_drivers/meta_instrument/Surface17_dependency_graph.py
MiniSean, line 1: Keep these changes local, the choice which cc-config is used is setup dependent.

pycqed/instrument_drivers/meta_instrument/HAL_Device.py
```diff
@@ -411,6 +411,10 @@ def measure_conditional_oscillation_multi(
         MC.set_sweep_points(p.sweep_points)
         d = self.get_int_avg_det()
         MC.set_detector_function(d)
+        
+        # Check low frequency qubit CZ amplitude
+        print("Amplituce SE", self.find_instrument("flux_lm_SE").vcz_amp_dac_at_11_02_NE())
```
MiniSean: No need to push this print statement

MiniSean (changes requested):
I have added some comments on which changes to keep locally and which to include in this merge.

msamiotis:
I have addressed most of the points made by @MiniSean .

MiniSean (approved):

